PR #1, opened 2016-10-05
Limin number of goroutines stuck on Sync syscall

Kubuxu:


whyrusleeping:
@Kubuxu does this have the intended effect of limiting thread creation?

Kubuxu:
It should have as Go scheduler creates thread when all other threads are blocked in syscalls and there is some work to do, blocking on a channel does not count as being blocked in syscall as thread blocked on channel can be separated from goroutine and reused for other work.

In short therms it should work, in a bit longer, I will test it.

whyrusleeping:
> TLDR; probably

Kubuxu:
I can confirm, after an add thread count jumps from 9 to 12 with the patch and to 62 without the patch.

sync.go
```diff
@@ -0,0 +1,40 @@
...
+	"os"
+	"runtime"
+)
+
+// don't block more than 16 threads on sync opearation
+// 16 should be able to sataurate most RAIDs
+// in case of two used disks per write (RAID 1, 5) and queue depth of 2,
+// 16 concurrent Sync calls should be able to saturate 16 HDDs RAID
+//TODO: benchmark it out, maybe provide tweak parmeter
+var syncSemaphore chan struct{} = make(chan struct{}, 16)
```
whyrusleeping: lets at least make this `16` value a constant.

sync.go
```diff
@@ -0,0 +1,40 @@
...
+
+func syncDir(dir string) error {
+	if runtime.GOOS == "windows" {
+		// dir sync on windows doesn't work: https://git.io/vPnCI
+		return nil
+	}
+
+	dirF, err := os.Open(dir)
```
whyrusleeping: any particular reason we open the dir before waiting for our token? That might result in a higher baseline fd count

Kubuxu: I wanted to minimize operations that wait for the token. In theory this operation could error out, so why wait for the token.

whyrusleeping: Fair enough

whyrusleeping:
@Kubuxu I'd love to see a test that stresses this a bit. Bonus points if you have a way to detect how many threads were spawned.

Kubuxu:
@whyrusleeping is crashing the test completely acceptable test? This is the only way I am able to come up with, using `debug.SetMaxThreads(x)`.

Kubuxu:
I am not able to write reliable tests for the sync.

whyrusleeping:
@Kubuxu Hrm... thats alright then.

PR #3, opened 2016-11-10
Improve Query Perforamance

kevina:
filepath.Walk() sorts the names and calls lstat() on each entry, both of which are unnecessary when all we need are the key names in random order.

Instead I just use the fact that the directory layout is fixed and just use Readdirnames.

With a sample datastore with 100000 small keys the speedup is around 2.2:
```
100000 ORIG 3443.543417 (+/- 63.378649) ms
100000 OPT 1558.430242 (+/- 28.740522) ms
```
When there are only 10000 keys the speedup is around 1.3:
```
10000 ORIG 199.053565 (+/- 12.956540) ms
10000 OPT 151.882705 (+/- 8.869852) ms
```

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -312,37 +312,60 @@ func (fs *Datastore) Query(q query.Query) (query.Results, error) {
 	reschan := make(chan query.Result, query.KeysOnlyBufSize)
 	go func() {
 		defer close(reschan)
-		err := filepath.Walk(fs.path, func(path string, info os.FileInfo, err error) error {
-			if os.IsNotExist(err) {
-				return nil
-			}
-			if err != nil {
-				log.Errorf("Walk func in Query got error: %v", err)
-				return err
-			}
-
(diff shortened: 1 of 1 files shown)


kevina:
I pushed the code for the benchmark on the kevina/faster-query-benchmark for lack of a better place.

kevina:
@whyrusleeping let me know if you have any questions

I am not sure how much will speedup AllKeysChan() in go-ipfs as there is some additional stuff going on but this is a good start in my opinion.

kevina:
I tested it with go-ipfs and `TEST_NO_FUSE=1 make test_short` passes.

whyrusleeping:
Ah, so this works because we essentially 'hard coding' the sharding depth. Got it.

My only concern is whether this works properly on windows. Do you have a windows computer you can test this on?

kevina:
@whyrusleeping I'm afraid I don't have easy access to a windows computer to test this; however, I don't see why it would not work.

whyrusleeping:
Tested on a windows cloud box, seems to work just fine there. LGTM

PR #6, opened 2016-12-07
Support using suffix of key for directory in addition to the prefix.

kevina:
Closes #6

PR #7, opened 2016-12-07
Ignore non-directories in the top level directory.

kevina:
Ignore non-directories in the top level directory to allow placing a README or similar file there.

This is based on #6.

### DIFF
--- a/flatfs_test.go
+++ b/flatfs_test.go
@@ -343,6 +343,11 @@ func testQuerySimple(dirFunc mkShardFunc, t *testing.T) {
 	temp, cleanup := tempdir(t)
 	defer cleanup()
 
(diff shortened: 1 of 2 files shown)


flatfs_test.go
```diff
@@ -343,6 +343,8 @@ func testQuerySimple(dirFunc mkShardFunc, t *testing.T) {
 	temp, cleanup := tempdir(t)
 	defer cleanup()
 
+	ioutil.WriteFile(filepath.Join(temp, "README"), []byte("something"), 0666)
```
whyrusleeping: might want to check that this doesnt error, just in case (if it errors and fails to create the file, then the test would pass regardless of the above fix)

kevina: Fixed now.

PR #9, opened 2016-12-08
Provide utility to convert from one Sharding type to another.

kevina:
This functionally is also provided as an exported function.

### DIFF
--- a/convert.go
+++ b/convert.go
@@ -110,14 +110,15 @@ func Move(oldPath string, newPath string, out io.Writer) error {
 		if err != nil {
 			return err
 		}
-		if inf.IsDir() || fn == "SHARDING" || fn == "_README" {
-			// if we are a director or generated file just remove it
+		if inf.IsDir() || fn == SHARDING_FN || fn == README_FN {
+			// we are an empty directory or generated file so just remove it
 			err := os.Remove(oldPath)
 			if err != nil {
 				return err
 			}
 		} else {
-			// else move it
+			// else we found something unexpected, so to be safe just move it
+			log.Warningf("found unexpected file in datastore directory: \"%s\", moving anyway", fn)
 			newPath := filepath.Join(newDS.path, fn)
 			err := osrename.Rename(oldPath, newPath)
 			if err != nil {
--- a/convert.go
+++ b/convert.go
@@ -1,0 +1,173 @@
+// Package flatfs is a Datastore implementation that stores all
+// objects in a two-level directory structure in the local file
+// system, regardless of the hierarchy of the keys.
+package flatfs
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+	"github.com/jbenet/go-os-rename"
+)
+
+func UpgradeV0toV1(path string, prefixLen int) error {
+	fun := Prefix(prefixLen)
+	err := WriteShardFunc(path, fun)
+	if err != nil {
+		return err
+	}
+	err = WriteReadme(path, fun)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func DowngradeV1toV0(path string) error {
+	err := os.Remove(filepath.Join(path, SHARDING_FN))
+	if err != nil {
+		return err
+	}
+	err = os.Remove(filepath.Join(path, README_FN))
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	return nil
+}
+
+func Move(oldPath string, newPath string, out io.Writer) error {
+	oldDS, err := Open(oldPath, false)
+	if err != nil {
+		return fmt.Errorf("%s: %v", oldPath, err)
+	}
+	newDS, err := Open(newPath, false)
+	if err != nil {
+		return fmt.Errorf("%s: %v", newPath, err)
+	}
+
+	if out != nil {
+		fmt.Fprintf(out, "Getting Keys...\n")
+	}
+	res, err := oldDS.Query(query.Query{KeysOnly: true})
+	if err != nil {
+		return err
+	}
+	entries, err := res.Rest()
+	if err != nil {
+		return err
+	}
+	prog := newProgress(len(entries), out)
+
+	if out != nil {
+		fmt.Fprintf(out, "Moving Keys...\n")
+	}
+
+	// first move the keys
+	for _, e := range entries {
+		err := moveKey(oldDS, newDS, datastore.RawKey(e.Key))
+		if err != nil {
+			return err
+		}
+		prog.Next()
+	}
+
+	if out != nil {
+		fmt.Fprintf(out, "\nCleaning Up...\n")
+	}
+
+	// now walk the old top-level directory
+	dir, err := os.Open(oldDS.path)
+	if err != nil {
+		return err
+	}
+	defer dir.Close()
+	names, err := dir.Readdirnames(-1)
+	if err != nil {
+		return err
+	}
+	for _, fn := range names {
+		if fn == "." || fn == ".." {
+			continue
+		}
+		oldPath := filepath.Join(oldDS.path, fn)
+		inf, err := os.Stat(oldPath)
+		if err != nil {
+			return err
+		}
+		if inf.IsDir() || fn == "SHARDING" || fn == "_README" {
+			// if we are a director or generated file just remove it
+			err := os.Remove(oldPath)
+			if err != nil {
+				return err
+			}
+		} else {
+			// else move it
+			newPath := filepath.Join(newDS.path, fn)
+			err := osrename.Rename(oldPath, newPath)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	if out != nil {
+		fmt.Fprintf(out, "All Done.\n")
+	}
+
+	return nil
+}
+
+func moveKey(oldDS *Datastore, newDS *Datastore, key datastore.Key) error {
+	_, oldPath := oldDS.encode(key)
+	dir, newPath := newDS.encode(key)
+	err := newDS.makeDirNoSync(dir)
+	if err != nil {
+		return err
+	}
+	err = osrename.Rename(oldPath, newPath)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+type progress struct {
+	total   int
+	current int
+
+	out io.Writer
+
+	start time.Time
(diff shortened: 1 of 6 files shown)


kevina:
This is ready for review.

kevina:
I also tested the utility on a real repo which is how I discovered https://github.com/ipfs/go-ipfs/issues/3463#issuecomment-265647047.  I also tested the reverse conversion.

convert.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if out != nil {
+		fmt.Fprintf(out, "Getting Keys...\n")
+	}
+	res, err := old.Query(query.Query{KeysOnly: true})
+	if err != nil {
+		return err
+	}
+	entries, err := res.Rest()
```
whyrusleeping: This will use a pretty hefty amount of memory. Why not just do things one at a time?

convert.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func moveKey(old *Datastore, new *Datastore, key datastore.Key) error {
+	_, oldPath := old.encode(key)
+	dir, newPath := new.encode(key)
+	err := new.makeDirNoSync(dir)
+	if err != nil {
+		return err
+	}
+	err = os.Rename(oldPath, newPath)
```
whyrusleeping: is `oldPath` a relative path? or is it prepended with the flatfs directory in encode?

whyrusleeping: Also, `os.Rename` may fail on windows, and if the source and destination are on different filesystems. We've been using go-os-rename to get around this

kevina: The idea is that source and destination should be on the same filesystem.   I was assuming this will fail on unix/OS X also.  The conversion operation will take a lot longer if a copy has to be made.

convert.go
kevina: @whyrusleeping I was copying the method you used in https://github.com/ipfs/fs-repo-migrations/blob/master/ipfs-3-to-4/migration/migration.go#L231.

I gather the keys first so I can get an accurate count for the progress bar.

And you are right it will take a good deal of memory, but it it didn't seam to cause anyone a problem the last time a conversion was made.

convert.go
kevina: > is oldPath a relative path? or is it prepended with the flatfs directory in encode?

It is prepended with the flatfs directory in encode?

convert.go
Kubuxu: It also means that no progress will be made during collection. Which might significantly extend length of the process as rename is very cheap operation in most filesystems.

kevina: For example using a ipfs repo I created on https://github.com/ipfs/go-ipfs/issues/3462 the time to gather the keys was around 2.5 minutes and it looks like the rename is going to take around 20 minutes.

whyrusleeping: My math might be off, but lets say i have a 4TB repo (which, it just so happens i do). Thats 4,000,000,000,000 bytes, which if we assume the average object size is (conservatively) 8k, thats half a billion keys. That means we're going to have an array of 500,000,000 query entries. Which, if i throw some random magic math at this, comes out to something like 30GB of memory used.

So yeah, it sucks to not have an idea of how many items there are going to be, but its really not acceptable to buffer these all in memory anymore (as i know we've done in the past).

Lets change the progress meter to just show total number of things processed so far, and skip the time estimates.

kevina: I'm fine with that.

kevina: @whyrusleeping Done.

kevina:
Note: based on top of #11. Can rebase if required once #11 is merged.

kevina:
@whyrusleeping I worked out the reason for the coverage decrease with @Kubuxu on IRC.  It was due to a large amount of error handling code.

kevina:
Note: I rebased this on #13.  Until #13 is merged you can compare the changes using https://github.com/ipfs/go-ds-flatfs/compare/kevina/SHARDING...kevina/convert.

I can rebase this on master if required once #13 is merged.

convert.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"gx/ipfs/QmRWDav6mzWseLWeYfVd5fvUKiVe9xNH29YfMF438fG364/go-datastore"
+	"gx/ipfs/QmRWDav6mzWseLWeYfVd5fvUKiVe9xNH29YfMF438fG364/go-datastore/query"
+	"gx/ipfs/QmaeRR9SpXumU5tYLRkq6x6pfMe8qKzxn4ujBpsTJ2zQG7/go-os-rename"
```
Kubuxu: It is rewritten.

kevina: Oops.  Sorry.  Fixed now.

convert.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+		if inf.IsDir() || fn == "SHARDING" || fn == "_README" {
+			// if we are a directory are generated file just remove it
+			err := os.Remove(oldPath)
+			if err != nil {
+				return err
+			}
+		} else {
+			// else move it
+			newPath := filepath.Join(new.path, fn)
```
whyrusleeping: what cases might end up here? Shouldnt we only expect to see the sharding and readme files?

kevina: Probably nothing, but if there is something in the old datastore I don't to preserve it.  The other option is to error out.  I don't want to remove the old datastore until everything is moved and the directory is empty.

convert.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/ipfs/go-datastore/query"
+	"github.com/jbenet/go-os-rename"
+)
+
+func Convert(oldPath string, oldShardFunc string, newPath string, newShardFunc string, out io.Writer) error {
+	old, err := New(oldPath, oldShardFunc, false)
+	if err != nil {
+		return err
+	}
+	new, err := New(newPath, newShardFunc, false)
```
Kubuxu: `new` is a keyword in Go, it can be confusing. How about `newDS` and `oldDS`.

convert.go
Kubuxu: Let's log it then.

convert.go
kevina: Will do. I am surprised that was even allowed.

kevina:
Note: This is going to need to be reworked a bit after #13 is merged.  I have the changes done but not pushed.

kevina:
Okay, I just reworked this as discussed over IRC.  I also provided a very simple utility and outlined the steps needed to convert from the old format to the new.

convert.go
```diff
@@ -0,0 +1,141 @@
...
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+	"github.com/jbenet/go-os-rename"
+)
+
+func UpgradeV0toV1(path string, prefixLen int) error {
```
whyrusleeping: We should check to make sure the datastore in question isnt already a V1

convert.go
```diff
@@ -0,0 +1,141 @@
...
+		if fn == "." || fn == ".." {
+			continue
+		}
+		oldPath := filepath.Join(oldDS.path, fn)
+		inf, err := os.Stat(oldPath)
+		if err != nil {
+			return err
+		}
+		if inf.IsDir() || fn == "SHARDING" || fn == "_README" {
+			// if we are a directory are generated file just remove it
```
whyrusleeping: s/are/or/

convert.go
```diff
@@ -0,0 +1,141 @@
...
+			err := os.Remove(oldPath)
+			if err != nil {
+				return err
+			}
+		} else {
+			// else move it
```
whyrusleeping: What sort of things are we expecting to be moving with this? There shouldnt be any other files, right?

flatfs/main.go
```diff
@@ -0,0 +1,96 @@
...
+	fmt.Fprintf(os.Stderr, "usage: %s create DIR SHARDFUN | upgrade DIR PREFIXLEN | downgrade DIR | move OLDDIR NEWDIR\n", os.Args[0])
+	os.Exit(1)
+}
+
+func fail(err error) {
+	fmt.Fprintf(os.Stderr, "%v\n", err)
+	os.Exit(1)
+}
+
+func main() {
```
whyrusleeping: oh i like this... In another PR can we perhaps get some more commands on this like query and 'get' ? That would be so cool

convert.go
kevina: I'm not expecting anything it is leftover from when the README was not auto-generated.  What do you want to do?  Error out.  Ignore it and then leave a directory that is not empty?

kevina:
Okay.  I will work on adding more tests.

whyrusleeping:
@kevina If the `Move` operation fails midway through, what is the correct way to recover from it?

convert.go
kevina: This is now done.  WriteShardFunc will fail if the SHARING file already exists.

kevina:
@whyrusleeping I added some tests to test failure conditions.

To recover from a failed move, just restart it.  See the tests.

convert.go
whyrusleeping: Unsure... how about  we keep things as is, but throw a warning saying we encountered an unexpected file?

convert_test.go
```diff
@@ -0,0 +1,234 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-ds-flatfs"
+	//"gx/ipfs/QmRWDav6mzWseLWeYfVd5fvUKiVe9xNH29YfMF438fG364/go-datastore/query"
+
+	rand "github.com/dustin/randbo"
```
Kubuxu: `rabdbo` has never fixed license problem, and same function is currently implemented in stdlib `math/rand.Read`.

kevina: This is not a new dependency.  It was already being used by `BenchmarkConsecutivePut` before any of my changes.  If this needs to be fixed it should probably be in a separate p.r.

Kubuxu: It isn't new dependency but PR introduces change that we know is not right. There is no cost in switching to `math/rand`.

kevina: My other P.R. also used `rabdbo`, which is why I found it odd that this was brought up now.  Anyway fixed.

Kubuxu: I might have not noticed that.

kevina: Okay.  I fixed that in https://github.com/ipfs/go-ds-flatfs/pull/14

PR #10, opened 2016-12-09
Add test to make sure suffix sharding prevents prefix clustering.

### LINKED ISSUE #8: Add a test to make sure suffix sharding prevents prefix clustering
Lets make sure to add a test that checks that adding a bunch of keys with a shared prefix doesnt result in all the entries landing in the same directory with suffixes chosen.

kevina:
Closes #8

### DIFF
--- a/flatfs_test.go
+++ b/flatfs_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base32"
 	"io/ioutil"
 	"math"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -14,7 +15,7 @@ import (
 	dstest "github.com/ipfs/go-datastore/test"
 	"github.com/ipfs/go-ds-flatfs"
 
-	rand "github.com/dustin/randbo"
+	randbo "github.com/dustin/randbo"
 )
 
 func tempdir(t testing.TB) (path string, cleanup func()) {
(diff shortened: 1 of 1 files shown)


kevina:
Note: based on top of #11.  Can rebase if required once #11 is merged.

flatfs_test.go
```diff
@@ -400,6 +401,50 @@ func testBatchDelete(dirFunc mkShardFunc, t *testing.T) {
...
+	tempdir, cleanup := tempdir(t)
+	defer cleanup()
+
+	fs, err := flatfs.New(tempdir, flatfs.NextToLast(1), false)
+	if err != nil {
+		t.Fatalf("New fail: %v\n", err)
+	}
+
+	r := rand.New()
+	N := 1024 // should be divisible by 32 so the math works out
```
Kubuxu: I would make it in for of `32*X` so math always works out.

flatfs_test.go
```diff
@@ -400,6 +401,50 @@ func testBatchDelete(dirFunc mkShardFunc, t *testing.T) {
...
+
+	dirs, err := ioutil.ReadDir(tempdir)
+	if err != nil {
+		t.Fatalf("ReadDir fail: %v\n", err)
+	}
+	if len(dirs) != 32 {
+		t.Fatalf("Expected 32 directories in %s", tempdir)
+	}
+	idealFilesPerDir := float64(N) / 32.0
+	tolerance := math.Floor(idealFilesPerDir * 0.50)
```
Kubuxu: That is 50% tolerance, right? Can we still reliably get closer than this?

We might also make the test bigger but make the seed static, as currently there is remote but non-zero chance of this test failing if stars align correctly.

flatfs_test.go
kevina: I prefer a number a direct number.  If we use 32* I will have to stop and do the math to figure the numbers of key used.  The comment also makes it explicit.  With just a 32*X then I would wonder why it is written like that.  @whyrusleeping what's your view.

flatfs_test.go
kevina: Okay I was able to decrease it to 20%.

PR #11, opened 2016-12-16
Add support for NextToLast shard function.

kevina:


flatfs_test.go
```diff
@@ -30,6 +30,11 @@ func tempdir(t testing.TB) (path string, cleanup func()) {
 	return path, cleanup
 }
 
+func tryAllShardFuncs(t *testing.T, testFunc func(mkShardFunc, *testing.T)) {
+	t.Run("prefix", func(t *testing.T) { testFunc(flatfs.Prefix, t) })
+	t.Run("suffix", func(t *testing.T) { testFunc(flatfs.Suffix, t) })
```
whyrusleeping: Should we put the `NextToLast` shardfunc here too?

kevina: Oops.  Thanks for catching this.  I tend to interactively commit changes to avoid commits the rewritten paths and missed this.

kevina: Fixed now.

PR #12, opened 2016-12-16
coverage: switch to codecov

Kubuxu:


kevina:
I'm all for this, I find the noise that coverall generate annoying. :)

PR #13, opened 2016-12-22
Store sharding function used in the repo.

kevina:
Store the sharding function used in the file "SHARDING" in the repo.
To make this work the sharding function is now always specified as a
string.

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -40,6 +40,8 @@ var _ datastore.Datastore = (*Datastore)(nil)
 
 type ShardFunc func(string) string
 
+var IPFS_DEF_SHARD = "v1/next-to-last/2"
+
 func New(path string, fun0 string, sync bool) (*Datastore, error) {
 	fun0 = NormalizeShardFunc(fun0)
 
--- a/flatfs.go
+++ b/flatfs.go
@@ -36,18 +36,15 @@ type Datastore struct {
 	sync bool
 }
 
-var _ datastore.Datastore = (*Datastore)(nil)
-
 type ShardFunc func(string) string
+
+var _ datastore.Datastore = (*Datastore)(nil)
 
 var (
 	ErrDatastoreExists       = errors.New("datastore already exist")
 	ErrDatastoreDoesNotExist = errors.New("datastore directory does not exist")
 	ErrShardingFileMissing   = fmt.Errorf("%s file not found in datastore", SHARDING_FN)
 )
-
-var IPFS_DEF_SHARD = NextToLast(2)
-var IPFS_DEF_SHARD_STR = IPFS_DEF_SHARD.String()
 
 func Create(path string, fun *ShardIdV1) error {
 
--- a/flatfs.go
+++ b/flatfs.go
@@ -53,21 +53,23 @@ func Create(path string, funStr string) error {
 		return err
 	}
 
+	fun, err := ParseShardFunc(funStr)
+	if err != nil {
+		return err
+	}
+
 	dsFun, err := ReadShardFunc(path)
 	switch err {
 	case ShardingFileMissing:
 		// fixme: make sure directory is empty and return an error if
 		// it is not
-		err := WriteShardFunc(path, funStr)
+		err := WriteShardFunc(path, fun)
 		if err != nil {
 			return err
 		}
-		return nil
+		err = WriteReadme(path, fun)
+		return err
 	case nil:
-		fun, err := ParseShardFunc(funStr)
-		if err != nil {
-			return err
-		}
 		if fun.String() != dsFun.String() {
 			return fmt.Errorf("specified shard func '%s' does not match repo shard func '%s'",
 				fun.String(), dsFun.String())
--- a/flatfs.go
+++ b/flatfs.go
@@ -46,7 +46,7 @@ var ShardingFileMissing = errors.New("SHARDING file not found in datastore")
 
 const IPFS_DEF_SHARD = "/repo/flatfs/shard/v1/next-to-last/2"
 
-func Create(path string, fun string) error {
+func Create(path string, funStr string) error {
 
 	err := os.Mkdir(path, 0777)
 	if err != nil && !os.IsExist(err) {
@@ -58,17 +58,19 @@ func Create(path string, fun string) error {
 	case ShardingFileMissing:
 		// fixme: make sure directory is empty and return an error if
 		// it is not
-		err := WriteShardFunc(path, fun)
+		err := WriteShardFunc(path, funStr)
 		if err != nil {
 			return err
 		}
 		return nil
 	case nil:
-		_ = dsFun
-		fun = NormalizeShardFunc(fun)
-		if fun != dsFun {
+		fun, err := ParseShardFunc(funStr)
+		if err != nil {
+			return err
+		}
+		if fun.String() != dsFun.String() {
 			return fmt.Errorf("specified shard func '%s' does not match repo shard func '%s'",
-				fun, dsFun)
+				fun.String(), dsFun.String())
 		}
 		return DatastoreExists
 	default:
@@ -84,20 +86,19 @@ func Open(path string, sync bool) (*Datastore, error) {
 		return nil, err
 	}
 
-	fun, err := ReadShardFunc(path)
+	shardId, err := ReadShardFunc(path)
 	if err != nil {
 		return nil, err
 	}
 
-	getDir, err := ShardFuncFromString(fun)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse shard func: %v", err)
 	}
 
 	fs := &Datastore{
 		path:      path,
-		shardFunc: fun,
-		getDir:    getDir,
+		shardFunc: shardId.String(),
+		getDir:    shardId.Func(),
 		sync:      sync,
 	}
 	return fs, nil
--- a/flatfs.go
+++ b/flatfs.go
@@ -43,7 +43,7 @@ type ShardFunc func(string) string
 var (
 	ErrDatastoreExists       = errors.New("datastore already exist")
 	ErrDatastoreDoesNotExist = errors.New("datastore directory does not exist")
-	ErrShardingFileMissing   = errors.New("SHARDING file not found in datastore")
+	ErrShardingFileMissing   = fmt.Errorf("%s file not found in datastore", SHARDING_FN)
 )
 
 const IPFS_DEF_SHARD = "/repo/flatfs/shard/v1/next-to-last/2"
@@ -68,7 +68,7 @@ func Create(path string, funStr string) error {
 			return err
 		}
 		if !isEmpty {
-			return fmt.Errorf("directory missing SHARDING file: %s", path)
+			return fmt.Errorf("directory missing %s file: %s", SHARDING_FN, path)
 		}
 
 		err = WriteShardFunc(path, fun)
--- a/flatfs.go
+++ b/flatfs.go
@@ -41,8 +41,8 @@ var _ datastore.Datastore = (*Datastore)(nil)
 type ShardFunc func(string) string
 
 var (
-	ErrDatastoreExists       = errors.New("Datastore already exist")
-	ErrDatastoreDoesNotExist = errors.New("Datastore directory does not exist")
+	ErrDatastoreExists       = errors.New("datastore already exist")
+	ErrDatastoreDoesNotExist = errors.New("datastore directory does not exist")
 	ErrShardingFileMissing   = errors.New("SHARDING file not found in datastore")
 )
 
@@ -63,9 +63,15 @@ func Create(path string, funStr string) error {
 	dsFun, err := ReadShardFunc(path)
 	switch err {
 	case ErrShardingFileMissing:
-		// fixme: make sure directory is empty and return an error if
-		// it is not
-		err := WriteShardFunc(path, fun)
+		isEmpty, err := DirIsEmpty(path)
+		if err != nil {
+			return err
+		}
+		if !isEmpty {
+			return fmt.Errorf("directory missing SHARDING file: %s", path)
+		}
+
+		err = WriteShardFunc(path, fun)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


whyrusleeping:
What are some example strings? We should make sure that these are somewhat nice multicodecs, cc @Kubuxu

flatfs.go
```diff
@@ -38,37 +40,52 @@ var _ datastore.Datastore = (*Datastore)(nil)
 
 type ShardFunc func(string) string
 
-func New(path string, getDir ShardFunc, sync bool) (*Datastore, error) {
-	fs := &Datastore{
-		path:   path,
-		getDir: getDir,
-		sync:   sync,
+func New(path string, fun0 string, sync bool) (*Datastore, error) {
```
kevina: Okay, but I will still need the "auto" test case as that is used by the conversion utility.

flatfs.go
```diff
@@ -38,37 +40,52 @@ var _ datastore.Datastore = (*Datastore)(nil)
...
 	}
-}
 
-func Suffix(suffixLen int) ShardFunc {
-	padding := strings.Repeat("_", suffixLen)
-	return func(noslash string) string {
-		str := padding + noslash
-		return str[len(str)-suffixLen:]
+	fun := ""
+	if fun0 == "auto" && fun1 == "auto" {
```
kevina: Okay.  I don't particular see the difference, but don't object either.  In the future I will try to remember that switch is generally preferred over multiple `else if`s.

flatfs.go
kevina: I didn't do this yet.  Do you basically want:
```
func New(path string, sync bool) (*Datastore, error) {
  NewWithFunc(path, "auto", sync)
}

func NewWithFunc(path string, fun0 string, sync bool) {...}
```

Is it that you don't like the special "auto" value?  I don't want to eliminate the use of that string as it makes writing the conversion function/utility easier, so using two separate functions won't really simplify any code.

Kubuxu:
What is the content of the SHARDING file directly? It isn't clear from the code?

kevina:
> What is the content of the SHARDING file directly? It isn't clear from the code?

Example file:
```
v1/next-to-last/2
```
As discussed in https://github.com/ipfs/go-ipfs/issues/3463

Kubuxu:
Could you change it to previously proposed: `/repo/flatfs/shard/v1/next-to-last/2`. There are no costs in it but it might make someone's day some time in future.

kevina:
@whyrusleeping agree with @Kubuxu? I have to problem doing it.

whyrusleeping:
Yeap, agreed with @Kubuxu

kevina:
> Could you change it to previously proposed: /repo/flatfs/shard/v1/next-to-last/2

Done.  Also refactored the parsing of the shard identifier string to make it more obvious what it going on.

flatfs.go
```diff
@@ -38,37 +40,53 @@ var _ datastore.Datastore = (*Datastore)(nil)
...
-	padding := strings.Repeat("_", prefixLen)
-	return func(noslash string) string {
-		return (noslash + padding)[:prefixLen]
+	fun := ""
+	switch {
+	case fun0 == "auto" && fun1 == "auto":
+		return nil, fmt.Errorf("shard function not specified")
+	case fun0 == "auto":
+		fun = fun1
+	case fun1 == "auto":
```
whyrusleeping: I think we should separate the task of opening an existing datastore and creating a new one. That way we can remove a lot of the weirdness here

kevina: I would rather not.  The "weirdness" is also there to allow opening an existing repo without the SHATDING file.

whyrusleeping: mmm.... We're going to have to run a migration anyways. I think thats acceptable.

kevina: @whyrusleeping, okay to do things the way you want I will basically have to redo this pull request and the conversion code will require some adjusting.  Let me know.

kevina: I started work on this.  Will have it ready some time Wed.

kevina:
@whyrusleeping okay please have another look, there are two fixme, but I think this is close to what you want.  There is a separate `Create` and `Open` method.

This fix also makes the flatfs a little more robust, espacally after the fixme to check if an existing directory without a SHARDING file is not empty.

shard.go
```diff
@@ -0,0 +1,138 @@
...
+	defer file.Close()
+	_, err = file.WriteString(str)
+	if err != nil {
+		return err
+	}
+	_, err = file.WriteString("\n")
+	if err != nil {
+		return err
+	}
+	if str == IPFS_DEF_SHARD {
```
kevina: @whyrusleeping I don't think returning an error is a good idea.  Especially if the flatfs datastore could be used outside of IPFS.  I also don't like the idea of a warning for the same reason, but can add one if you still think it is a good idea.

I will separate the step from the writing of the sharding function.

flatfs_test.go
```diff
@@ -214,40 +218,42 @@ func testStorage(p *params, t *testing.T) {
 	if !seen {
 		t.Error("did not see the data file")
 	}
+	if fs.ShardFunc() == flatfs.IPFS_DEF_SHARD && !haveREADME {
+		t.Error("expected _README file")
+	} else if fs.ShardFunc() == flatfs.IPFS_DEF_SHARD && !haveREADME {
```
kevina: Oops.  Yes.  And I have a bug.  Will fix.

shard.go
```diff
@@ -0,0 +1,138 @@
...
+func parseShardFunc(str string) shardId {
+	str = strings.TrimSpace(str)
+	parts := strings.Split(str, "/")
+	// ignore prefix for now
+	if len(parts) > 3 {
+		parts = parts[len(parts)-3:]
+	}
+	switch len(parts) {
+	case 3:
+		return shardId{version: parts[0], funName: parts[1], param: parts[2]}
```
kevina: Okay, I will see what I can do.  I originally had parseShardFunc return an error but decided to push all the error checking into Func().  Will try refactoring again and see if I can make it work better the way I think you want.

kevina: Fixed now.

kevina:
@whyrusleeping I should have addressed all your concerns in the latest review except for returning an error if no README can be created because I am not sure that it was a good idea.  See my comment.

kevina:
Note the two special files are named:
```
  _README
  SHARDING
```
I use '_README' because that is what @jbenet wanted.  For consistency I can also rename 'SHARDING' to '_SHARDING' if you want.  Let me know.  For the record, I think using the special name '_README' is a good idea to help make it clear there is something special about it.

flatfs.go
```diff
@@ -38,37 +40,86 @@ var _ datastore.Datastore = (*Datastore)(nil)
 
 type ShardFunc func(string) string
 
-func New(path string, getDir ShardFunc, sync bool) (*Datastore, error) {
-	fs := &Datastore{
-		path:   path,
-		getDir: getDir,
-		sync:   sync,
+var DatastoreExists = errors.New("Datastore already exist")
+var DatastoreDoesNotExist = errors.New("Datastore directory does not exist")
```
Kubuxu: Please use one var block and `Err` name prefix.

shard.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("empty shard identifier")
+	}
+	if str[0] == '/' {
+		if !strings.HasPrefix(str, PREFIX) {
+			return nil, fmt.Errorf("invalid prefix in shard identifier: %s", str)
+		}
+		str = str[len(PREFIX):]
+	}
+	parts := strings.Split(str, "/")
+	if len(parts) == 3 {
```
Kubuxu: Is the v1 optional? it doesn't have to be. Let's make it simpler.

shard.go
```diff
@@ -0,0 +1,133 @@
...
+		id.fun = NextToLast(param)
+	default:
+		return nil, fmt.Errorf("expected 'prefix', 'suffix' or 'next-to-last' got: %s", id.funName)
+	}
+
+	return id, nil
+}
+
+func ReadShardFunc(dir string) (*ShardIdV1, error) {
+	buf, err := ioutil.ReadFile(filepath.Join(dir, "SHARDING"))
```
Kubuxu: Let's make the name of the file constant as it is reused.

flatfs.go
```diff
@@ -38,37 +40,86 @@ var _ datastore.Datastore = (*Datastore)(nil)
...
+	} else if err != nil {
+		return nil, err
+	}
+
+	shardId, err := ReadShardFunc(path)
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
```
Kubuxu: Double err handler.

shard.go
```diff
@@ -0,0 +1,133 @@
...
+	// ignore prefix for now
+	if len(str) == 0 {
+		return nil, fmt.Errorf("empty shard identifier")
+	}
+	if str[0] == '/' {
+		if !strings.HasPrefix(str, PREFIX) {
+			return nil, fmt.Errorf("invalid prefix in shard identifier: %s", str)
+		}
+		str = str[len(PREFIX):]
+	}
```
Kubuxu: Let's use TrimPrefix, and check if output string is different than the input, if it is then prefix was trimmed.

shard.go
```diff
@@ -0,0 +1,133 @@
...
+		err := ioutil.WriteFile(filepath.Join(dir, "_README"), []byte(IPFS_DEF_SHARD), 0444)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func Prefix(prefixLen int) ShardFunc {
+	padding := strings.Repeat("_", prefixLen)
```
Kubuxu: It there a better way to do it? It seems clever but it might have performance penalty (string concat), and it is done for every key, potentially multiple time.

Kubuxu: This applies to other ShardFuncs too.

kevina: This has not changed.  The function was simply moved and copied.  Yeah there might be a better way, but I don't want to address it in this pull request.

flatfs.go
kevina: Oops.  Will fix.

flatfs.go
kevina: Okay.

shard.go
kevina: There might be yes.  But this code has not changed, just moved.  The same thing was also done before any of my shading changes.  So for now (at least this p.r.) I would rather leave it alone.

shard.go
kevina: I don't agree (with "it doesn't have to be") but @whyrusleeping would probably say the same thing so I will change it.

shard.go
kevina: Okay.

shard.go
kevina: Do we have a policy on this?  In this particular case it seams like an overkill to me.  We also don't seam to be very consistent about it elsewhere.

whyrusleeping: We try to use constants for these sorts of things when we can, There are definitely cases where we forget to do so, but in general its a good idea

kevina: Fine, Done.  However, I think it makes code harder to read and I also am horrible at naming the things.

shard.go
kevina: There might be yes.  But this code has not changed, just moved.  The same thing was also done before any of my shading changes.  So for now (at least this p.r.) I would rather leave it alone.

[Sorry if this is a duplicate.  GitHub is being annoying.]

kevina:
@Kubuxu @whyrusleeping I should have addressed all your concerns.  I also implemented the fixme.

The only outstanding issue is returning an error if no README can be created because I am not sure that it was a good idea.  From my comment above:

> > Lets separate the step of writing the readme from writing the sharding function, and also make sure to error out or print a log message if we try to write a readme for an unsupported sharding type

> @whyrusleeping I don't think returning an error is a good idea. Especially if the flatfs datastore could be used outside of IPFS. I also don't like the idea of a warning for the same reason, but can add one if you still think it is a good idea.

shard.go
```diff
@@ -0,0 +1,135 @@
...
+	_, err = file.WriteString("\n")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func WriteReadme(dir string, id *ShardIdV1) error {
+	if id.String() == IPFS_DEF_SHARD {
+		err := ioutil.WriteFile(filepath.Join(dir, README_FN), []byte(IPFS_DEF_SHARD), 0444)
```
Kubuxu: It should be `README_DEF_SHARD`

shard.go
```diff
@@ -0,0 +1,135 @@
...
+	defer file.Close()
+	_, err = file.WriteString(id.String())
+	if err != nil {
+		return err
+	}
+	_, err = file.WriteString("\n")
+	if err != nil {
+		return err
+	}
+	return nil
```
Kubuxu: All above can be just `return err`.

flatfs.go
```diff
@@ -38,37 +40,90 @@ var _ datastore.Datastore = (*Datastore)(nil)
...
+
+		err = WriteShardFunc(path, fun)
+		if err != nil {
+			return err
+		}
+		err = WriteReadme(path, fun)
+		if err != nil {
+			return err
+		}
+		return nil
```
Kubuxu: can be just `return err` too (no need for if).

flatfs.go
```diff
@@ -38,37 +40,90 @@ var _ datastore.Datastore = (*Datastore)(nil)
...
+	ErrShardingFileMissing   = fmt.Errorf("%s file not found in datastore", SHARDING_FN)
+)
+
+const IPFS_DEF_SHARD = "/repo/flatfs/shard/v1/next-to-last/2"
+
+func Create(path string, funStr string) error {
+
+	err := os.Mkdir(path, 0777)
+	if err != nil && !os.IsExist(err) {
+		return nil
```
Kubuxu: Shouldn't it be `return err`?

kevina: Yes.

shard.go
kevina: Okay.  Told you I was bad a naming things. :)

kevina: Oops.  Misunderstood and didn't see my own error.  Will fix.

kevina:
@Kubuxu anything else :)

kevina:
Note: Just rebased to clean up the commits, nothing else has changed.

kevina:
Note: I just pushed some more commits to avoid the use of strings wherever possible.

PR #14, opened 2017-01-19
Eliminate github.com/dustin/randbo dependency.

kevina:
'rabdbo' has never fixed license problem, and same function is currently implemented in stdlib 'math/rand.Read'.


kevina:
Looks like @whyrusleeping already did this.  Making a new p.r. with a cleanup.

PR #16, opened 2017-02-12
Remove unfinished write transactions

Kubuxu:
This needs to be updated in fs-repo-migrations

### DIFF
--- a/convert.go
+++ b/convert.go
@@ -150,7 +150,7 @@ func Move(oldPath string, newPath string, out io.Writer) error {
 			}
 		} else {
 			// else we found something unexpected, so to be safe just move it
-			log.Warningf("found unexpected file in datastore directory: \"%s\", moving anyway", fn)
+			log.Warningf("found unexpected file in datastore directory: \"%s\", moving anyway\n", fn)
 			newPath := filepath.Join(newDS.path, fn)
 			err := osrename.Rename(oldPath, newPath)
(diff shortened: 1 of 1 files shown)


convert.go
```diff
@@ -110,8 +112,38 @@ func Move(oldPath string, newPath string, out io.Writer) error {
...
+
+			names, err := indir.Readdirnames(-1)
+			if err != nil {
+				return err
+			}
+
+			for _, n := range names {
+				p := filepath.Join(oldPath, n)
+				// part of unfinished write transaction
+				// remove it
```
kevina: I am not 100% convinced this is a good thing to do.  The file could contain important data.  I could be wrong but this seams like the type of thing that would end up in the "lost+found" directory if this where a traditional Unix filesystem.  Maybe we could move the file into into a special directory.  @whyrusleeping what do you think.

kevina:
@Kubuxu could you do me a favor and fix this line:

> log.Warningf("found unexpected file in datastore directory: \"%s\", moving anyway", fn)

so that there is a newline at the end?

convert.go
whyrusleeping: My original suggestion was to delete it. I still think that is the right thing to do. The write that created this file failed, giving the user no expectation of the data existing anywhere on disk.

kevina: Okay, I'm fine with that.

PR #18, opened 2017-02-17
convert: close the directory explicitly, don't use defer

kevina:


Kubuxu:
There is still one, line 101.

PR #19, opened 2017-04-29
improve permissions on files created with flatfs

whyrusleeping:
cc @kyledrake

### DIFF
--- a/util.go
+++ b/util.go
@@ -1,8 +1,12 @@
 package flatfs
 
 import (
-	"os"
 	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+	"sync"
+	"time"
 )
 
 // From: http://stackoverflow.com/questions/30697324/how-to-check-if-directory-on-path-is-empty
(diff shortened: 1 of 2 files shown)


kyledrake:
Context: go-ipfs uses tempfile to create and then moves it to the final position, but in the current configuration uses 600 for permissions. Umask apparently will only lower permissions from what was initially set, so setting the tempfiles to 0666 should allow the user to control the default permissions with the umask (http://stackoverflow.com/a/23842371).

:+1: thank you!

util.go
```diff
@@ -19,3 +23,49 @@ func DirIsEmpty(name string) (bool, error) {
...
+// Copied from io/ioutil.TempFile, changed default permissions to 0640 instead of 0600
+func TempFile(dir, prefix string) (f *os.File, err error) {
+	if dir == "" {
+		dir = os.TempDir()
+	}
+
+	nconflict := 0
+	for i := 0; i < 10000; i++ {
+		name := filepath.Join(dir, prefix+nextSuffix())
+		f, err = os.OpenFile(name, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0640)
```
kevina: Why 0640 instead of the 0666 as suggested on stack overflow?

whyrusleeping: because i hadnt yet pushed the fix

kevina:
@whyrusleeping I would think that you need to include the go copyright somewhere.  I would put all the code copied from ioutil/tempfile.go into its own file and include the go copyright at the top.

Issue #22, opened 2017-11-05
Unable to handle keys with namespaces


dignifiedquire:
I came across this when trying to put something with a key like `"/foo/bar"` this will result in a panic, as only the `foo` directory is created but not the `bar`. 

Digging deeper the reason is that `encode` does not handle namespaces in keys properly. 

This is my current test output which shows how the sharding function messes up the namespaced key:

```
=== RUN   TestPutGetNamespaces
=== RUN   TestPutGetNamespaces/prefix
key: /q/u/u/x
getDir: q/
noslash: q/u/u/x
=== RUN   TestPutGetNamespaces/suffix
key: /q/u/u/x
getDir: /x
noslash: q/u/u/x
=== RUN   TestPutGetNamespaces/next-to-last
key: /q/u/u/x
getDir: u/
noslash: q/u/u/x
--- FAIL: TestPutGetNamespaces (0.00s)
    --- FAIL: TestPutGetNamespaces/prefix (0.00s)
    	flatfs_test.go:88: Put fail: rename /var/folders/v2/4jf047qs3vvgz1b8ty6khx6c0000gn/T/test-datastore-flatfs-652809544/q/put-518678023 /var/folders/v2/4jf047qs3vvgz1b8ty6khx6c0000gn/T/test-datastore-flatfs-652809544/q/q/u/u/x.data: no such file or directory
    --- FAIL: TestPutGetNamespaces/suffix (0.00s)
    	flatfs_test.go:88: Put fail: rename /var/folders/v2/4jf047qs3vvgz1b8ty6khx6c0000gn/T/test-datastore-flatfs-073166778/x/put-852409297 /var/folders/v2/4jf047qs3vvgz1b8ty6khx6c0000gn/T/test-datastore-flatfs-073166778/x/q/u/u/x.data: no such file or directory
    --- FAIL: TestPutGetNamespaces/next-to-last (0.00s)
    	flatfs_test.go:88: Put fail: rename /var/folders/v2/4jf047qs3vvgz1b8ty6khx6c0000gn/T/test-datastore-flatfs-267857660/u/put-194386475 /var/folders/v2/4jf047qs3vvgz1b8ty6khx6c0000gn/T/test-datastore-flatfs-267857660/u/q/u/u/x.data: no such file or directory
```

Kubuxu:
[...]

PR #23, opened 2017-11-05
Add failing test for namespaced keys

dignifiedquire:
Ref #22

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -116,8 +116,11 @@ func (fs *Datastore) ShardStr() string {
 }
 
 func (fs *Datastore) encode(key datastore.Key) (dir, file string) {
+	fmt.Printf("key: %s\n", key.String())
 	noslash := key.String()[1:]
 	dir = filepath.Join(fs.path, fs.getDir(noslash))
+	fmt.Printf("getDir: %s\n", fs.getDir(noslash))
+	fmt.Printf("noslash: %s\n", noslash)
 	file = filepath.Join(dir, noslash+extension)
 	return dir, file
 }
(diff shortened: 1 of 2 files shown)


flatfs.go
```diff
@@ -116,8 +116,11 @@ func (fs *Datastore) ShardStr() string {
 }
 
 func (fs *Datastore) encode(key datastore.Key) (dir, file string) {
+	fmt.Printf("key: %s\n", key.String())
```
Stebalien: These look like debugging print statements.

flatfs_test.go
```diff
@@ -71,6 +71,39 @@ func testPut(dirFunc mkShardFunc, t *testing.T) {
...
+	if !ok {
+		t.Fatalf("expected []byte from Get, got %T: %v", data, data)
+	}
+
+	if g, e := string(buf), input; g != e {
+		t.Fatalf("Get gave wrong content: %q != %q", g, e)
+	}
+}
+
+func TestPutGetNamespaces(t *testing.T) { tryAllShardFuncs(t, testPutGetNamespaces) }
```
Stebalien: One line functions are generally not idiomatic go. Try running `go fix ./...` and/or `go fmt ./...` on your repo to fix stuff like this.

flatfs.go
dignifiedquire: Yep they are, until we figured out how to fix the issue.

flatfs_test.go
dignifiedquire: I just copy and pasted from the above test, to keep the original style.

dignifiedquire: formatting cleanup should be probably done in a different commit/pr

PR #26, opened 2017-11-26
Make Datastore.walk() error more specific

lgierth:
This one is currently showing up in little bursts on the gateways and coule be a bit more specific :)

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -432,7 +432,7 @@ func (fs *Datastore) walk(path string, reschan chan query.Result) error {
 
 		key, ok := fs.decode(fn)
 		if !ok {
-			log.Warning("failed to decode entry in flatfs")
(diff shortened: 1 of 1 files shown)


flatfs.go
```diff
@@ -432,7 +432,7 @@ func (fs *Datastore) walk(path string, reschan chan query.Result) error {
 
 		key, ok := fs.decode(fn)
 		if !ok {
-			log.Warning("failed to decode entry in flatfs")
+			log.Warningf("failed to decode flatfs entry: %s", key)
 			continue
```
kevina: `fn` should be used instead of `key` for the warning.  If there was an error `key` may be empty.

lgierth: D'oh -- I forgot to commit and push that last night after testing it with Gx :) Thanks.

PR #27, opened 2018-02-16
Feat: Implement a PersistentDatastore by adding DiskUsage method

hsanjuan:
This adds DiskUsage().

This datastore would have a big performance hit if we walked the
filesystem to calculate disk usage everytime.

Therefore I have opted to keep tabs of current disk usage by
walking the filesystem once during "Open" and then adding/subtracting
file sizes on Put/Delete operations.

On the plus:
  * Small perf impact
  * Always up to date values
  * No chance that race conditions will leave DiskUsage with wrong values

On the minus:
  * Slower Open() - it run Stat() on all files in the datastore
  * Size does not match real size if a directory grows large
    (at least on ext4 systems). We don't track directory-size changes,
    only use the creation size.

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -181,6 +181,8 @@ func (fs *Datastore) makeDirNoSync(dir string) error {
 
 var putMaxRetries = 6
 
+// Put stores a key/value in the datastore. Note that it does not replace
+// already existing values. If you wish to replace a value, do a Delete first.
 func (fs *Datastore) Put(key datastore.Key, value interface{}) error {
 	val, ok := value.([]byte)
 	if !ok {
--- a/flatfs.go
+++ b/flatfs.go
@@ -264,7 +264,7 @@ func (fs *Datastore) renameAndUpdateDiskUsage(tmpPath, path string) error {
 	// Rename and add new file's diskUsage. If the rename fails,
 	// it will either a) Re-add the size of an existing file, which
 	// was sustracted before b) Add 0 if there is no existing file.
-	err := osrename.Rename(tmpPath, path)
+	err = osrename.Rename(tmpPath, path)
 	fs.updateDiskUsage(path, true)
 	return err
 }
@@ -598,20 +598,33 @@ func (fs *Datastore) calculateDiskUsage() error {
 
 	// Otherwise walk the datastore files
 	var du int64
-	err := filepath.Walk(fs.path,
-		func(path string, info os.FileInfo, err error) error {
-			if err != nil {
-				return err
+	dirs := []string{fs.path}
+	i := 0
+
+	for {
+		if i >= len(dirs) {
+			break
+		}
+
+		dir := dirs[i]
+		i++
+
+		f, err := os.Open(dir)
+
+		if err != nil {
+			return err
+		}
+		stats, err := f.Readdir(0) // 0 means no limit
+		if err != nil {
+			return err
+		}
+		for _, fi := range stats {
+			du += fi.Size()
+			if fi.IsDir() {
+				dirs = append(dirs, filepath.Join(dir, fi.Name()))
 			}
-			du += info.Size()
-			return nil
-		})
-
-	if err != nil {
-		log.Errorf("cannot calculate disk usage: %s", err)
-		return err
+		}
 	}
-
 	atomic.StoreInt64(&fs.diskUsage, du)
 	return nil
 }
--- a/flatfs.go
+++ b/flatfs.go
@@ -240,8 +240,16 @@ func (fs *Datastore) doPut(key datastore.Key, val []byte) error {
 	}
 	closed = true
 
-	err = osrename.Rename(tmp.Name(), path)
-	if err != nil {
+	if err := os.Link(tmp.Name(), path); err != nil {
+		if os.IsExist(err) {
+			// Exit cleanly. Note: deferred function will remove
+			// the temp file and we don't need sync.
+			return nil
+		}
+		return err
+	}
+
+	if err := os.Remove(tmp.Name()); err != nil {
 		return err
 	}
 	removed = true
@@ -317,7 +325,16 @@ func (fs *Datastore) putMany(data map[datastore.Key]interface{}) error {
 
 	// move files to their proper places
 	for fi, path := range files {
-		if err := osrename.Rename(fi.Name(), path); err != nil {
+		if err := os.Link(fi.Name(), path); err != nil {
+			if os.IsExist(err) {
+				// Deferred function will take care of
+				// removing temp file
+				continue
+			}
+			return err
+		}
+
+		if err := os.Remove(fi.Name()); err != nil {
 			return err
 		}
 
@@ -372,12 +389,11 @@ func (fs *Datastore) Has(key datastore.Key) (exists bool, err error) {
 func (fs *Datastore) Delete(key datastore.Key) error {
 	_, path := fs.encode(key)
 
-	// when removing an unexisting file
-	// this does nothing.
-	fs.updateDiskUsage(path, false)
+	fSize := fileSize(path)
 
 	switch err := os.Remove(path); {
 	case err == nil:
+		atomic.AddInt64(&fs.diskUsage, -fSize)
 		return nil
 	case os.IsNotExist(err):
 		return datastore.ErrNotFound
--- a/flatfs.go
+++ b/flatfs.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -25,6 +26,7 @@ var log = logging.Logger("flatfs")
 const (
 	extension      = ".data"
 	diskUsageChCap = 1024
+	diskUsageFile  = "diskUsage.cache"
 )
 
 type Datastore struct {
@@ -436,9 +438,17 @@ func (fs *Datastore) walkTopLevel(path string, reschan chan query.Result) error 
 	return nil
 }
 
+// calculateDiskUsage tries to read the diskUsageFile for a cached
+// diskUsage value, otherwise walks the datastore files.
 func (fs *Datastore) calculateDiskUsage() error {
-	var du int64
+	// Try to obtain a previously stored value from disk
+	if persDu := fs.readDiskUsageFile(); persDu > 0 {
+		fs.diskUsageCh <- int64(persDu)
+		return nil
+	}
 
+	// Otherwise walk the datastore files
+	var du int64
 	err := filepath.Walk(fs.path,
 		func(path string, info os.FileInfo, err error) error {
 			if err != nil {
@@ -488,6 +498,33 @@ func (fs *Datastore) updateDiskUsage(path string, add bool) {
 	if fsize != 0 {
 		fs.diskUsageCh <- fsize
 	}
+}
+
+func (fs *Datastore) persistDiskUsageFile() {
+	// Store DiskUsage on clean shutdowns.
+	du, err := fs.DiskUsage()
+	if err != nil {
+		// do not store on errors. just ignore them
+		return
+	}
+
+	duB := []byte(fmt.Sprintf("%d", du))
+	ioutil.WriteFile(filepath.Join(fs.path, diskUsageFile), duB, 0644)
+}
+
+// deletes the diskUsageFile after reading
+func (fs *Datastore) readDiskUsageFile() uint64 {
+	fpath := filepath.Join(fs.path, diskUsageFile)
+	defer os.Remove(fpath)
+	duB, err := ioutil.ReadFile(fpath)
+	if err != nil {
+		return 0
+	}
+	i, err := strconv.ParseUint(string(duB), 10, 64)
+	if err != nil {
+		return 0
+	}
+	return i
 }
 
 // DiskUsage implements the PersistentDatastore interface
@@ -551,6 +588,7 @@ func (fs *Datastore) walk(path string, reschan chan query.Result) error {
 
 func (fs *Datastore) Close() error {
 	close(fs.diskUsageCh)
+	fs.persistDiskUsageFile()
 	return nil
 }
 
--- a/flatfs.go
+++ b/flatfs.go
@@ -58,7 +58,7 @@ const (
 var _ datastore.Datastore = (*Datastore)(nil)
 
 var (
-	ErrDatastoreExists       = errors.New("datastore already exist")
+	ErrDatastoreExists       = errors.New("datastore already exists")
 	ErrDatastoreDoesNotExist = errors.New("datastore directory does not exist")
 	ErrShardingFileMissing   = fmt.Errorf("%s file not found in datastore", SHARDING_FN)
 )
--- a/flatfs.go
+++ b/flatfs.go
@@ -173,9 +173,7 @@ func Open(path string, syncFiles bool) (*Datastore, error) {
 		getDir:    shardId.Func(),
 		sync:      syncFiles,
 		diskUsage: 0,
-		opMap: &opMap{
-			ops: sync.Map{},
-		},
+		opMap:     new(opMap),
 	}
 
 	// This sets diskUsage to the correct value
@@ -263,7 +261,9 @@ func (fs *Datastore) renameAndUpdateDiskUsage(tmpPath, path string) error {
 		return err
 	}
 
-	// Rename and add new file's diskUsage
+	// Rename and add new file's diskUsage. If the rename fails,
+	// it will either a) Re-add the size of an existing file, which
+	// was sustracted before b) Add 0 if there is no existing file.
 	err := osrename.Rename(tmpPath, path)
 	fs.updateDiskUsage(path, true)
 	return err
--- a/flatfs.go
+++ b/flatfs.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/ipfs/go-datastore"
@@ -22,7 +23,8 @@ import (
 var log = logging.Logger("flatfs")
 
 const (
-	extension = ".data"
+	extension      = ".data"
+	diskUsageChCap = 1024
 )
 
 type Datastore struct {
@@ -33,6 +35,9 @@ type Datastore struct {
 
 	// sychronize all writes and directory changes for added safety
 	sync bool
+
+	diskUsage   uint64
+	diskUsageCh chan int64
 }
 
 type ShardFunc func(string) string
@@ -94,10 +99,24 @@ func Open(path string, sync bool) (*Datastore, error) {
 	}
 
 	fs := &Datastore{
-		path:     path,
-		shardStr: shardId.String(),
-		getDir:   shardId.Func(),
-		sync:     sync,
+		path:        path,
+		shardStr:    shardId.String(),
+		getDir:      shardId.Func(),
+		sync:        sync,
+		diskUsage:   0,
+		diskUsageCh: make(chan int64, diskUsageChCap),
+	}
+
+	go fs.diskUsageUpdater()
+	// This sets diskUsage to the correct value
+	// It might be slow, but allowing it to happen
+	// while the datastore is usable might
+	// cause diskUsage to not be accurate.
+	err = fs.calculateDiskUsage()
+	if err != nil {
+		// Cannot stat() all
+		// elements in the datastore.
+		return nil, err
 	}
 	return fs, nil
 }
@@ -154,7 +173,11 @@ func (fs *Datastore) makeDirNoSync(dir string) error {
 		if !os.IsExist(err) {
 			return err
 		}
+		return nil
 	}
+
+	// Track DiskUsage of this NEW folder
+	fs.updateDiskUsage(dir, true)
 	return nil
 }
 
@@ -224,6 +247,9 @@ func (fs *Datastore) doPut(key datastore.Key, val []byte) error {
 		return err
 	}
 	removed = true
+
+	// Update diskUsage
+	fs.updateDiskUsage(path, true)
 
 	if fs.sync {
 		if err := syncDir(dir); err != nil {
@@ -300,6 +326,9 @@ func (fs *Datastore) putMany(data map[datastore.Key]interface{}) error {
 
 		// signify removed
 		ops[fi] = 2
+
+		// Track disk usage
+		fs.updateDiskUsage(path, true)
 	}
 
 	// now sync the dirs for those files
@@ -346,6 +375,9 @@ func (fs *Datastore) Has(key datastore.Key) (exists bool, err error) {
 
 func (fs *Datastore) Delete(key datastore.Key) error {
 	_, path := fs.encode(key)
+
+	fs.updateDiskUsage(path, false)
+
 	switch err := os.Remove(path); {
 	case err == nil:
 		return nil
@@ -404,6 +436,78 @@ func (fs *Datastore) walkTopLevel(path string, reschan chan query.Result) error 
 	return nil
 }
 
+func (fs *Datastore) calculateDiskUsage() error {
+	var du int64
+
+	err := filepath.Walk(fs.path,
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			du += info.Size()
+			return nil
+		})
+
+	if err != nil {
+		log.Errorf("cannot calculate disk usage: %s", err)
+		return err
+	}
+
+	fs.diskUsageCh <- du
+	return nil
+}
+
+// diskUsageUpdater should run in a goroutine.
+// only call after calculateDiskUsage() has stored
+// a correct value in fs.diskUsage.
+func (fs *Datastore) diskUsageUpdater() {
+	var curDu int64
+
+	for duInc := range fs.diskUsageCh {
+		curDu += duInc // duInc might be < 0
+		atomic.StoreUint64(&fs.diskUsage, uint64(curDu))
+	}
+}
+
+func fileSize(path string) int64 {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return 0
+	}
+	return fi.Size()
+}
+
+// updateDiskUsage reads the size of path and sends it either
+// as an increment or a decrement to the diskUsageCh.
+func (fs *Datastore) updateDiskUsage(path string, add bool) {
+	fsize := fileSize(path)
+	if !add {
+		fsize = -fsize
+	}
+
+	if fsize != 0 {
+		fs.diskUsageCh <- fsize
+	}
+}
+
+// DiskUsage implements the PersistentDatastore interface
+// and returns the current disk usage in bytes used by
+// this datastore.
+//
+// The size is approximative and may slightly differ from
+// the real disk values.
+func (fs *Datastore) DiskUsage() (uint64, error) {
+	// it may differ from real disk values if
+	// the filesystem has allocated for blocks
+	// for a directory because it has many files in it
+	// we don't account for "resized" directories.
+	// In a large datastore, the differences should be
+	// are negligible though.
+
+	du := atomic.LoadUint64(&fs.diskUsage)
+	return du, nil
+}
+
 func (fs *Datastore) walk(path string, reschan chan query.Result) error {
 	dir, err := os.Open(path)
 	if err != nil {
@@ -446,6 +550,7 @@ func (fs *Datastore) walk(path string, reschan chan query.Result) error {
 }
 
 func (fs *Datastore) Close() error {
+	close(fs.diskUsageCh)
 	return nil
 }
 
--- a/flatfs.go
+++ b/flatfs.go
@@ -246,7 +246,6 @@ func (fs *Datastore) doPut(key datastore.Key, val []byte) error {
 	}
 	removed = true
 
-	// Update diskUsage
 	fs.updateDiskUsage(path, true)
 
 	if fs.sync {
@@ -325,7 +324,6 @@ func (fs *Datastore) putMany(data map[datastore.Key]interface{}) error {
 		// signify removed
 		ops[fi] = 2
 
-		// Track disk usage
 		fs.updateDiskUsage(path, true)
 	}
 
@@ -374,6 +372,8 @@ func (fs *Datastore) Has(key datastore.Key) (exists bool, err error) {
 func (fs *Datastore) Delete(key datastore.Key) error {
 	_, path := fs.encode(key)
 
+	// when removing an unexisting file
+	// this does nothing.
 	fs.updateDiskUsage(path, false)
 
 	switch err := os.Remove(path); {
--- a/flatfs.go
+++ b/flatfs.go
@@ -27,8 +27,18 @@ var log = logging.Logger("flatfs")
 
 const (
 	extension                  = ".data"
-	DiskUsageFile              = "diskUsage.cache"
 	diskUsageCheckpointPercent = 1.0
+)
+
+var (
+	// DiskUsageFile is the name of the file to cache the size of the
+	// datastore in disk
+	DiskUsageFile = "diskUsage.cache"
+	// DiskUsageFilesAverage is the number of files per folder
+	// to stat in order to calculate the size of the datastore.
+	// Any other files will be considered to have the average size
+	// of the previously processed ones
+	DiskUsageFilesAverage = 100
 )
 
 const (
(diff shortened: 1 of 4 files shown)


lgierth:
> Slower Open() - it run Stat() on all files in the datastore

That's a bummer for the gateways -- can't have a full repo scan on every restart there.

Can we make the bookkeeping persistent so that it survives restart?

hsanjuan:
@lgierth is it so bad? In my gateway tests a while ago it did take a few seconds, but it's once during boot and while ipfs is not hammering the disk at the same time.

That said, I can leave note of the size on a file, sure, and only re-scan when not present.

hsanjuan:
@lgierth there you go

Stebalien:
There are two downsides to caching disk usage:

1. Downgrade, do something, upgrade (as we haven't bumped the repo version).
2. It'll be even more out-of-date directory sizes (meh?).

An alternative is to calculate the initial size lazily. The first call to DiskUsage will be slow but it won't halt the world.

hsanjuan:
Without looking at `Walk` internals, walking lazily probably suffers from race conditions where a file may be counted twice (one through Put as soon as we Open the datastore and the other one as the lazy walk finally gets to its folder). Also, the other way around with in the case of delete.

So it also has some downsides... (and the fact that it may return really low sizes at the beginning which may be to whoever is looking at that metric).

I lean towards caching even though it's not perfect...

Stebalien:
> walking lazily probably suffers from race conditions where a file may be counted twice

Hm. You're right. We'd have to track which parts we've walked and which parts we haven't. That would be *horrible*.

Yeah, this sounds like the best solution. Note: If we want a more accurate solution, we could cache the size of the files and recompute the sizes of the directories but it's probably not worth it.

flatfs.go
```diff
@@ -33,6 +37,9 @@ type Datastore struct {
 
 	// sychronize all writes and directory changes for added safety
 	sync bool
+
+	diskUsage   uint64
+	diskUsageCh chan int64
```
Stebalien: Why not just use `atomic.AddInt64` (and store it as an int64)?

flatfs.go
```diff
@@ -404,6 +438,113 @@ func (fs *Datastore) walkTopLevel(path string, reschan chan query.Result) error
...
+func (fs *Datastore) diskUsageUpdater() {
+	var curDu int64
+
+	for duInc := range fs.diskUsageCh {
+		curDu += duInc // duInc might be < 0
+		atomic.StoreUint64(&fs.diskUsage, uint64(curDu))
+	}
+}
+
+func fileSize(path string) int64 {
```
Stebalien: We could actually be *slightly* more accurate by adding in the size of the filename.

flatfs.go
```diff
@@ -404,6 +438,113 @@ func (fs *Datastore) walkTopLevel(path string, reschan chan query.Result) error
...
+func (fs *Datastore) persistDiskUsageFile() {
+	// Store DiskUsage on clean shutdowns.
+	du, err := fs.DiskUsage()
+	if err != nil {
+		// do not store on errors. just ignore them
+		return
+	}
+
+	duB := []byte(fmt.Sprintf("%d", du))
+	ioutil.WriteFile(filepath.Join(fs.path, diskUsageFile), duB, 0644)
```
Stebalien: Should really use a tempfile with a rename.

hsanjuan:
> probably not worth it.

agreed. Will you be so kind to greenlight the code or leave feedback @Stebalien ?

flatfs.go
hsanjuan: you think? in my tests `du` on the disk matched the values we produce here. aren't filenames in ext* part of the directory entry, and the directory entry counts as fulll 4096 bytes block even if empty. So im not sure if adding filenames is double-counting some bytes..

flatfs.go
hsanjuan: can do, can you explain the logic behind it though?

hsanjuan: btw, my /tmp folder lives in a partition different from my ipfs storage. In the case when you have a system with an ssd and a spinning disk, it is not unusual that /tmp /var etc are mounted on the spinning disk (to reduce the number of writes to the ssd) and the rest of things on the ssd, so imagine the performance hit of current implementation with tmp files and renames..

flatfs.go
Stebalien: I was trying to account for the case where we add a bunch of files without re-stating the directories (in which case, they'll be much larger than 4096). However, that case may not really be worth considering. If it comes to it, we can always consider recomputing once in a while.

flatfs.go
Stebalien: Ah, no. I mean an `ioutil.TempFile` as we do in `doPut`. That is, create a temporary file, write to it, close it, and then atomically rename it into place (so we don't end up with a half-written or corrupted file if we crash while recording the size). Probably not much of an issue in this case but still good practice.

flatfs.go
hsanjuan: yes, indeed..

flatfs.go
hsanjuan: yes. i just realized that tmpfiles are created in the same folder. yeah makes sense. Thanks!

flatfs.go
hsanjuan: ok, leaving this as it is. To my usecase, it doesn't really matter. We can improve it when we need to.

Stebalien: Fair enough.

Stebalien:
So, reasoning through this carefully, there are a few annoying race conditions:

* If two users *write* the same block at the same time, we'll double-count the file.
* Same for two simultaneous deletes.
* A simultaneous write + delete could end up not recording the fact that we've freed disk space.

For the delete case, we should:

1. Compute the disk space.
2. Remove the file.
3. Update the usage iff *we* removed the file (the call to Remove succeeded).

For the write case... I think we'll have to use `os.Link(old, new)` and `os.Remove(old)`. That should work on all operating systems and should avoid overwriting the file. Note: this assumes that we never want to *replace* values in datastores. If we do, we'd have to do... we could do a remove and replace dance...

There are a few os-specific ways to make this faster but cross-platform low-level filesystem stuff always kind of sucks.

hsanjuan:
@Stebalien you're right.

I have followed your indications. I'm not super happy though:

* Not allowing overwrite of values seems like a high price to pay and changes datastore behavior
* I cannot get my test of concurrent mixed put+delete over the same key to pass with `sync = false`. Sure, deviation is small, but it seems some sort of race exists still at a lower level. I'm happy ignoring this.
* I have tried to do the `remove and replace` dance. This introduces further race conditions (cannot remove and replace atomically) and I'm not sure it's possible to do that dance without making things horrible or recurring to `sync.Mutex`..

My feelings overall:

* DiskUsage bookeeping should not deeply impact datastore performance.
* DiskUsage should be more or less accurate for the common usage of this datastore, but it doesn't need to exact.

Thus, if we can live with the non-overwrite limitation i'd leave things like this, otherwise, I'd return to the buggy behaviour before (fixing the delete part, which is easy). What do you think?

Stebalien:
> My feelings overall:

Overall, I agree. Performance is already poor enough as it is.

> Thus, if we can live with the non-overwrite limitation i'd leave things like this, otherwise, I'd return to the buggy behaviour before (fixing the delete part, which is easy). What do you think?

My worry is that simultaneous writes/deletes are probably quite common and, if they're not perfectly balanced, the estimated disk usage will skew over time. We might, actually, just want to use a synchronization mechanism. I wouldn't have a *global* lock, but we can track in-progress puts/deletes on a per-file basis. That is, I'd have a map (either a `sync.Map` or a map protected by a mutex) mapping file names to some form of synchronization object. Every operation would keep retrying itself until it either (a) observes itself fail or (b) observes some concurrent operation (or itself) succeed.

That is, given a timeline like:

```
<--op 1---> // takes the "lock"
  <--op 2-> // waits
   <-op 3-> // waits
```

If op1 succeeds, we can serialize these operations as `<op2><op3><op1>` (and say that `op2` and `op3` are valid for no observable time and therefor optimize them out :smile:). If op1 fails, we'll need to retry at least one of the other operations (in practice, they'll both retry and start at the top).

This should actually be *faster* than the existing implementation given concurrent operations and should be just as "correct".

Does this sound too complicated? Any better ideas?

hsanjuan:
@Stebalien ok, I have implemented the approach you suggested. Works well for my tests (added an extra one too for putMany()). Hopefully we can settle here :)

The tricky code should be sufficiently commented, but let me know if something's not clear.

Note, Mac tests fail because `too many open files`. I'll tune that later.

flatfs.go
```diff
@@ -22,9 +25,31 @@ import (
 var log = logging.Logger("flatfs")
 
 const (
-	extension = ".data"
+	extension     = ".data"
+	diskUsageFile = "diskUsage.cache"
 )
 
+const (
+	putOp = iota
```
Kubuxu: IMO use named type of `opT`.

flatfs.go
```diff
@@ -33,6 +58,13 @@ type Datastore struct {
 
 	// sychronize all writes and directory changes for added safety
 	sync bool
+
+	diskUsage int64
+
+	// opMap handles concurrent write operations (put/delete)
+	// to the same key
+	opMap     map[string]*sync.Mutex
```
Kubuxu: It might be better to use concurrent map. It should be essentially free when we are doing concurrent operations on different keys.

Stebalien: :+1: although I think the OS will be the bottleneck here.

Stebalien:
> In my opinion it is too high of a price to pay for a feature that might not be used. Having it does lazily, on request and reporting an error when the scan has not yet finished would be far better.

We've discussed this but it's prohibitively complicated as we need a snapshot size. One solution would be to disable writes until the scan is complete but that complicates matters and isn't really a solution.

flatfs.go
```diff
@@ -183,7 +266,89 @@ func (fs *Datastore) Put(key datastore.Key, value interface{}) error {
...
+		fs.opMap[keyStr] = opLock
+	}
+	fs.opMapLock.Unlock()
+
+	// Attempt to run the operation.
+	// Only one operation can run at a time
+	// on the same key. We use opLock
+	// for that.
+	opLock.Lock()
+	defer func() {
```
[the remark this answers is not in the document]
hsanjuan: ops, left over from some debugging

hsanjuan:
> We've discussed this but it's prohibitively complicated as we need a snapshot size. One solution would be to disable writes until the scan is complete but that complicates matters and isn't really a solution.

We could write the DiskUsage-file regularly (say when disk usage difference from last checkpoint >1%) so that we can read that on start. It may not be 100% accurate in the case of a crash, but it might be close enough for our needs.

Kubuxu:
@hsanjuan SGWM, it also seems to me that triggering full re-scan on `ipfs repo fsck` would be good feature to have. This way if DiskUsage file diverged too far, fsck would fix it.

Any idea how to nicely wire it in?

hsanjuan:
@Kubuxu that will need interface support in go-datastore. Should we have an `Fsck()` method for `PersistentDatastore` that does this (and potentially other things if needed)?

Kubuxu:
I am thinking if PersistentDatastore is only one needing this and it seems so. I also recall Bager needing hook after GC was performed so it could perform its own internal GC. I don't know if it was solved  (@magik6k).

flatfs.go
```diff
@@ -183,7 +266,89 @@ func (fs *Datastore) Put(key datastore.Key, value interface{}) error {
...
+	err := fs.doOp(oper)
+	if err != nil {
+		// We failed. Tell the user. Leave the operation
+		// in opMap so that other potentially concurrent operations
+		// can try.
+		// Note: opMap will contain keys for operations which just
+		// failed, even if they did not compete with another
+		// one. If datastore operations fail all the time
+		// opMap will grow as it retains all keys of failed operations
+		// and only deletes them on success.
```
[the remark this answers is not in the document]
hsanjuan: Ok, let me try this

magik6k:
The GC interface is wired into go-ipfs master already (merged in https://github.com/ipfs/go-ipfs/pull/4578)

I created 2 more interfaces in go-datastore when making the GC one - https://github.com/ipfs/go-datastore/blob/master/datastore.go#L88-L103. `Check` from `CheckedDatastore` is meant to be triggered on `ipfs repo fsck`. I'm not sure where `Scrub` should get invoked, probably separate command like `ipfs repo scrub`.

`ScrubbedDatastore.Scrub` is probably where we want to wire this into. (`Check` / `repo fsck` should be fast)

flatfs.go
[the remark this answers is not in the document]
hsanjuan: @Stebalien the `delete` in `Finish` causes panics with the map (fixed by locking/unlocking around it). I have implemented the whole thing with a `sync.Map` instead.

whyrusleeping:
Could we get some benchmarks on how this will affect the overall performance of the datastore?

flatfs.go
```diff
@@ -404,6 +589,127 @@ func (fs *Datastore) walkTopLevel(path string, reschan chan query.Result) error
...
+		fsize = -fsize
+	}
+
+	if fsize != 0 {
+		newDu := atomic.AddInt64(&fs.diskUsage, fsize)
+		fs.checkpointDiskUsage(newDu)
+	}
+}
+
+func (fs *Datastore) checkpointDiskUsage(newDuInt int64) {
```
Stebalien: There *is* a serious drawback to this. We could write a bunch of data but crash before checkpointing. The old way was *much* safer because we didn't leave the file on disk while it could be invalid.

flatfs.go
```diff
@@ -154,12 +199,46 @@ func (fs *Datastore) makeDirNoSync(dir string) error {
...
+}
+
+// This function always runs under an opLock. Therefore, only one thread is
+// touching the affected files.
+func (fs *Datastore) renameAndUpdateDiskUsage(tmpPath, path string) error {
+	fi, err := os.Stat(path)
+
+	// Destination exists, we need to discount it from diskUsage
+	if fs != nil && err == nil {
+		atomic.AddInt64(&fs.diskUsage, -fi.Size())
```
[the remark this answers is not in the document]
Stebalien: This will still need to be addressed (the remove could fail).

Kubuxu:
From my tests, `ipfs repo stat` on Biham (9TB of data in go-ipfs) didn't finish after 13h, I've cancelled it. I will leave `du` on blockstore running over the night. `du` is probably as close to optimal size scanning as we will get. I expect it also to take at least hours. I don't think that we can allow for this kind "semi random" (after unexpected circumstances) delays in starting of ipfs node.

flatfs.go
hsanjuan: You would need to crash precisely in the write operation which would do a checkpoint while having written a really large entry (with respects to the current size) for the checkpoint to be really skewed from actual size. I think it's not so much of a big deal as soon as the size of the datastore is sufficiently big to the size of the entries written to it.

hsanjuan:
> I don't think that we can allow for this kind "semi random" (after unexpected circumstances) delays in starting of ipfs node.

Ideally this only happens once, if the checkpoint file is not there (which should be in subsequent starts).

It might be that stat-walking large repos is too expensive even to do it once. We could be smart and list entries in folders and multiply by 256K. This would give an approximate ballpark in a hopefully way faster way. I'm happy sacrificing accuracy as in a large-file context most blocks will probably be full.

Kubuxu:
The `du` scan finished after 4h. Yet the `ipfs repo stat` previously didn't finish after 13h.

flatfs.go
```diff
@@ -404,6 +587,127 @@ func (fs *Datastore) walkTopLevel(path string, reschan chan query.Result) error
...
+func (fs *Datastore) calculateDiskUsage() error {
+	// Try to obtain a previously stored value from disk
+	if persDu := fs.readDiskUsageFile(); persDu > 0 {
+		atomic.StoreInt64(&fs.diskUsage, persDu)
+		return nil
+	}
+
+	// Otherwise walk the datastore files
+	var du int64
+	err := filepath.Walk(fs.path,
```
kevina: I would not use filepath.Walk here.  It sorts the directory names which is unnecessary and might lead to excessive seeking as stat is called on each directory entry.  It could also be doing other unnecessary things.  There is no reason this can't as efferent as `du`.  I would use `Readdir` directly.

flatfs.go
Stebalien: Fair enough.

flatfs.go
```diff
@@ -94,10 +168,25 @@ func Open(path string, sync bool) (*Datastore, error) {
...
-		shardStr: shardId.String(),
-		getDir:   shardId.Func(),
-		sync:     sync,
+		path:      path,
+		shardStr:  shardId.String(),
+		getDir:    shardId.Func(),
+		sync:      syncFiles,
+		diskUsage: 0,
+		opMap: &opMap{
+			ops: sync.Map{},
```
Stebalien: No need to manually initialize this. `opMap: new(opMap)` is sufficient. Same with `diskUsage` (go initializes to "zero value").

flatfs.go
```diff
@@ -154,12 +243,44 @@ func (fs *Datastore) makeDirNoSync(dir string) error {
...
+	fi, err := os.Stat(path)
+
+	// Destination exists, we need to discount it from diskUsage
+	if fs != nil && err == nil {
+		atomic.AddInt64(&fs.diskUsage, -fi.Size())
+	} else if !os.IsNotExist(err) {
+		return err
+	}
+
+	// Rename and add new file's diskUsage
```
Stebalien: Should probably explain why we update disk usage regardless.

hsanjuan:
> Currently, this will sometimes return an error when deleting something that isn't there and sometimes won't

Any operation that performs a delete and gets success can assume that they deleted something. I don't think it's too important that the caller knows that it was not its own delete but a concurrent one that actually really deleted the file. Whenever no error is returned, you can assume something WAS there and something WAS deleted "not long ago".

hsanjuan:
@kevina using `Readdir` now.

kevina:
@Kubuxu can you see if using `Readdir` helps performace?

hsanjuan:
@whyrusleeping @Stebalien my feeling here is that performance depends mostly on the OS and disk, and our changes (other than on a cold start), won't impact so much (and might improve the scenarios with concurrent writes to the same keys).

Let's have a look at existing benchmarks, which use random keys and non concurrent Puts, and sync = false.

### Using /dev/shm (in memory)

New:

```
goos: linux
goarch: amd64
pkg: github.com/ipfs/go-ds-flatfs
BenchmarkConsecutivePut-8          20000           1307640 ns/op
BenchmarkBatchedPut-8              20000           1358376 ns/op
PASS
ok      github.com/ipfs/go-ds-flatfs    84.285s
```

Master:

```
goos: linux
goarch: amd64
pkg: github.com/ipfs/go-ds-flatfs
BenchmarkConsecutivePut-8          20000           1240002 ns/op
BenchmarkBatchedPut-8              20000           1721207 ns/op
PASS
ok      github.com/ipfs/go-ds-flatfs    88.808s
```

Master again:

```
goos: linux
goarch: amd64
pkg: github.com/ipfs/go-ds-flatfs
BenchmarkConsecutivePut-8          20000           1838413 ns/op
BenchmarkBatchedPut-8              20000           1510638 ns/op
PASS
ok      github.com/ipfs/go-ds-flatfs    96.605s
```

New again:

```
goos: linux
goarch: amd64
pkg: github.com/ipfs/go-ds-flatfs
BenchmarkConsecutivePut-8          20000           1083198 ns/op
BenchmarkBatchedPut-8              20000           1120730 ns/op
PASS
ok      github.com/ipfs/go-ds-flatfs    75.054s
```

They don't seem to show anything terrible going on. If anything, things got faster **which I can't explain**. This happens consistently even after I repeat the benchmark several times.

---

### Testing with /tmp (spinning disk)

New:

```
goos: linux
goarch: amd64
pkg: github.com/ipfs/go-ds-flatfs
BenchmarkConsecutivePut-8           3000            575120 ns/op
BenchmarkBatchedPut-8               5000           1446841 ns/op
PASS
ok      github.com/ipfs/go-ds-flatfs    17.183s
```

Master:

```
goos: linux
goarch: amd64
pkg: github.com/ipfs/go-ds-flatfs
BenchmarkConsecutivePut-8          10000           2407771 ns/op
BenchmarkBatchedPut-8               5000           8187870 ns/op
PASS
ok      github.com/ipfs/go-ds-flatfs    189.733s
```

Master again:

```
goos: linux
goarch: amd64
pkg: github.com/ipfs/go-ds-flatfs
BenchmarkConsecutivePut-8          10000           3104441 ns/op
BenchmarkBatchedPut-8              10000           3186439 ns/op
PASS
ok      github.com/ipfs/go-ds-flatfs    82.227s
```

New again:

```
goos: linux
goarch: amd64
pkg: github.com/ipfs/go-ds-flatfs
BenchmarkConsecutivePut-8           5000           2775705 ns/op
BenchmarkBatchedPut-8               5000           2180260 ns/op
PASS
ok      github.com/ipfs/go-ds-flatfs    32.512s

```

It would seem that the benchmarks try to adjust for more iterations in master (which means it was faster for lower N?). But iterating more eventually becomes a huge penalty so master is slower per loop.

---

From these benchmarks I can't see a negative impact of this branch over performance of `Put`. I think we are mostly subject to OS and Disk induced delays when interacting with the storage.

hsanjuan:
@kevina @Kubuxu running a "quick" test now

hsanjuan:
@Kubuxu @kevina : quick test in `biham` has taken about 8 hours. I don't think there's too much overhead in `Walk` vs `Readdir`. The stat call in `du` might be optimized for size only when in go it fetches a bunch of stuff.

I think it might be best to provide a fast-calc option which assumes 256K file sizes when directories have too many files in them. We just can't take 8 hours (even with a lazy implementation).

whyrusleeping:
How in the world does this make things faster? I'm not understanding that...

kevina:
@hsanjuan using something like strace might revival why `du` may be faster, it could be that go is making needed system calls to get extra information that is not needed just to determine the size of the file.

magik6k:
GNU du uses `fts_open` instead of readdir - https://github.com/coreutils/coreutils/blob/master/src/du.c#L684.

> I think it might be best to provide a fast-calc option which assumes 256K file sizes when directories have too many files in them. We just can't take 8 hours (even with a lazy implementation).

Since all the data is hashed, it will be evenly distributed in most scenarios (though this assumption breaks when we introduce identity hashes and is only relevant to use of the datastore in ipfs) - we could take a random sample of sizes - few K - and multiply the average size of the sampled blocks by block count. Assuming even block distribution this should give reasonably good approximation of the data volume without taking ages to compute.

flatfs.go
```diff
@@ -22,9 +26,39 @@ import (
...
+const (
+	opPut = iota
+	opDelete
+	opRename
+)
+
+var _ datastore.Datastore = (*Datastore)(nil)
+
+var (
+	ErrDatastoreExists       = errors.New("datastore already exist")
```
magik6k: Typo, should be `exists`

hsanjuan:
I have taken a stab at disk usage stimation. I have run some tests in biham.

* `Readdirnames` get's very very very slow after a few dozens of folders have been read, even though the amount of files in them is more or less the same (around 35000). This sucks because it's not so much the `stat` part that was causing all the delays but the directory listing. I'm not sure why.
* Because of this, I cannot do `Readdirnames` for all of the prefix folders because things get very slow even if I don't stat all the files in them.
* I therefore only read the 100 first files in folders (including the base one) and take that and calculate the average. In other words, I read 100 block files from the first 100 prefix-folders and extrapolate that too figure out the size of all files in all folders.
* This takes a few seconds in biham (`<10 secs`, acceptable by any means).
* Estimated size in biham is : `7438083607446` vs real size from the other day test (with slow 8h-long method): `7536515831332`. Seems somewhere between 1%-2% deviation.

Comments? I would like to take larger samples but `Readdirnames` tends to get really slow and right now I don't see an obvious reason. Perhaps I'm doing something stupid.

flatfs.go
```diff
@@ -404,6 +597,172 @@ func (fs *Datastore) walkTopLevel(path string, reschan chan query.Result) error
...
+	if err != nil {
+		return 0, err
+	}
+
+	totalFiles := len(files)
+	filesToProcess := files
+	nonProcessed := 0
+	if DiskUsageFilesAverage < totalFiles && DiskUsageFilesAverage > 0 {
+		nonProcessed = totalFiles - DiskUsageFilesAverage
+		filesToProcess = files[0:DiskUsageFilesAverage]
```
magik6k: We should pick those randomly, as the order in which the files are returned may correspond to the order they were added in (this as quite system-specific, but may happen in theory), and thus won't get us the uniform size distribution.

kevina:
> Readdirnames get's very very very slow after a few dozens of folders have been read, even though the amount of files in them is more or less the same (around 35000). This sucks because it's not so much the stat part that was causing all the delays but the directory listing. I'm not sure why.

This is weird, i am guessing it is a memory or GC problem.  Is the same slow down seen when `Query` (which calls `walkTopLevel` and `walk`) is used?

Stebalien:
> Since all the data is hashed, it will be evenly distributed in most scenarios (though this assumption breaks when we introduce identity hashes and is only relevant to use of the datastore in ipfs)

It also breaks due to different codecs (if we sort, that is). Blocks with one codec will be sorted before another.

> we could take a random sample of sizes - few K - and multiply the average size of the sampled blocks by block count. Assuming even block distribution this should give reasonably good approximation of the data volume without taking ages to compute.

Regardless, this scheme is still valid (assuming a random sample).

> I therefore only read the 100 first files in folders (including the base one) and take that and calculate the average. In other words, I read 100 block files from the first 100 prefix-folders and extrapolate that too figure out the size of all files in all folders

This *should* be fine given the sharding function we use. However, it will break the second we start storing blobs with names that don't end in a hash. I'd rather stat 100 random files out of 100 random folders. Still not perfect but better.

Stebalien:
> > Currently, this will sometimes return an error when deleting something that isn't there and sometimes won't

> Any operation that performs a delete and gets success can assume that they deleted something. I don't think it's too important that the caller knows that it was not its own delete but a concurrent one that actually really deleted the file. Whenever no error is returned, you can assume something WAS there and something WAS deleted "not long ago".

Unless I'm mistaken, it will *also* return a success when we both delete and put something at the same time even if the put wins and there was nothing there before.

flatfs.go
```diff
@@ -55,6 +56,10 @@ var (
 	ErrShardingFileMissing   = fmt.Errorf("%s file not found in datastore", SHARDING_FN)
 )
 
+func init() {
+	rand.Seed(time.Now().UTC().UnixNano())
```
Kubuxu: I would switch to package local `Rand` instance as you can't guarantee that `Seed` won't be called with other calls to `Rand` concurrently.

hsanjuan: I don' t understand what you mean

hsanjuan: `Seed, unlike the Rand.Seed method, is safe for concurrent use. ` https://golang.org/pkg/math/rand/#Seed

hsanjuan:
Progress and tests so far:

* It selects random files/folders when the limit is less than the number of them
* It has a folder limit because the slowest operation is readdirnames and that happens on folders, so I figure out if we read less folders we can compensate stat-ing more files.
* Readdirnames is very slow on `biham` when the folders are not cached (so it was fast because they had been accessed earlier). With current implementation it will take 3 minutes to start the first time after this change comes in. It reads 50 random folders (of around 1000) and uses 100 for the files in them.
* This gives a 5% error in the calculated size. More accuracy will mean more time (what's acceptable?). But ***this only happens on the first start when there is no a diskUsage.cache file***.

Are we happy with this?

flatfs.go
Kubuxu: I've missed it, sorry. I was looking at Rand.Seed.

Stebalien:
Still need to fix #30 but we can do that in a subsequent PR.

kevina:
Okay there seams to be a lot going on here.  Please give me a day or two too look over this.

Note that I don't think identity hashes will cause any problems with sampling because if we implement them as I envisioned the data stores won't have to worry about them because they won't even be stored in the datastore (as there is no need).

hsanjuan:
Thanks @kevina , will wait

kevina:
If we are going to estimate the size I think instead of taking a fixed sample size we should iterate until we get an acceptable level of accuracy.  The accuracy can be estimated using the formula for the _standard error for finite population_ (the _finite population correction factor_ is the important part, otherwise the error won't shrink as we take larger samples).  We can continue to take samples until this number with within an acceptable margin (I would say 1%).

Since we are not even reading all the directories we have two estimate numbers we are working with (1) The estimate number of files and (2) the estimate mean file size.  The standard error needs to be computed independently for each and then property combined to produce a final estimate.

The population size for (1) is the number of directories.  The population size for (2) is the number of files.  Since the population size for (2) is only an estimate, this might need to be taken into account when calculating the correction factor.

I would prefer that DiskUsage() is exact.  If it is not, I think it is important we have a high confidence on the accuracy of the estimate, we can't get this confidence by simply taking a fixed sample size and calling it good.

kevina:
I should also note that as far as I know standard error assumes a normal distribution which may not be the case in terms of block sizes, in particular there are likely to be a lot of very small and an possible even a larger number of 256K blocks.

I still think we should iterate, I am just not 100% sure when to know to stop.

flatfs.go
```diff
@@ -22,9 +27,50 @@ import (
...
+	// The size of the rest of the files in a folder will be assumed
+	// to be the average of the values obtained. This includes
+	// regular files and directories.
+	DiskUsageFilesAverage = 100
+	// DiskUsageFoldersAverage is the maximum number of folders
+	// to read in order to calculate the size of the datastore. Note that
+	// folders are also considered files with respect to
+	// DiskUsageFilesAverage (therefore this value only makes sense
+	// if it's lower).
+	DiskUsageFoldersAverage = 50
```
kevina: This is not a good threshold, even modest datastores will have a large number of directories.  We should only consider skipping folders after there are a certain number of files in a directory and reading them truly becomes a problem.

hsanjuan:
> We should limit sampling to huge datastores where computing the size will take hours. For most datastores (of maybe say 100 GB of less) this calculation should only take a few minutes and we should provide an exact number.

I can already tell you that, based on my tests, getting within 1% of error and being sure about it during sampling is going to take too long.

What I can do instead is to set a time limit rather than a folder limit and read all random samples possible (I'll play with larger number of files per folder) until the acceptable time limit is hit (say 5 minutes). This should give accuracy with small datastores (or those equipped with fast disks) and do our best on large/slow ones. Also, simplicity.

Would this make you happy enough @kevina ?

flatfs.go
Kubuxu: @kevina the directories contain random sample of the whole datastore (due to how prefix is selected) so it isn't necessary to sample many of them. In theory if the datastore is big enough we could sample just random files from one directory.

kevina:
@Kubuxu @hsanjuan, I have a problem with `DiskUsage()` being approximate. The documentation for DiskUsage says `DiskUsage returns the space used by a datastore, in bytes.`.  At very least it is bad UX and could cause problems.  Worse is the fact that the user has no way to know if it is an appropriation.

Perhaps a better way forward is if it is not possible to exactly calculate the DiskUsage() in a reasonable time, we should leave leave it undefined and provide a tool to calculate the DiskUsage or to optionally make an estimate and update the value based on a command line flag.  This way the user is at least aware that it is an approximation.  They can also schedule to run the command at a convenient time.

In order to estimate if we can calculate the DiskUsage() in a reasonable time suggestion we sample a few random directories and count the number of files, if the count is above a threshold we give up and somehow mark the datastore as having an undefined DiskUsage().  We can also use a timeout.

Sorry, but I just can't in good consonance approve this P.R. as written.

flatfs.go
kevina: @Kubuxu that is true, but if the datastore is small, this will still be a approximation and not an exact value.

kevina:
However, if people can show me other examples where DiskUsage calculation for a Database is an approximation, I would more okay with it.  We would still need to update the documentation to make it clear that this is not an exact number.

magik6k:
As far as I am concerned the badger implementation of Size doesn't return the exact size on disk (at least not for every file-system) as doing so with the many specifics of it would be quite hard.

The problem the DiskUsage method is trying to solve is to provide a fast way to get the size of data stored within ipfs datastore.

One use case is described in https://github.com/ipfs/go-ipfs/pull/4550, the other is to display repository size in the desktop app where this was a problem too. For both of those cases being within .1% (or even 1%) should be perfectly acceptable imo. Having this method take more than few seconds is not as it makes it unusable for those use cases.

Yes, we should document that the result may be an approximation, but making return 'undefined' when it would take too long doesn't solve ether of the use cases above and just adds a feature which can't be used in most realistic scenarios (most datastores where you'd care about size will be too big to calculate exact size in a reasonable time)

kevina:
> One use case is described in ipfs/go-ipfs#4550, the other is to display repository size in the desktop app where this was a problem too. For both of those cases being within .1% (or even 1%) should be perfectly acceptable imo. Having this method take more than few seconds is not as it makes it unusable for those use cases.

(1) Based on the latest estimates the accuracy will likely be more like 5%, if it was 1% (and with reasonable confidence we know its 1% I would object less)

(2) If I understand things correctly this method will only take long when the method has not been called before.  On new datastores (after this lands) and after this method is called the first time, the DiskUsage will be kept in sync.  Thus my proposal is if it takes too long in it initial calculation, don't do it and leave it up to the user to update manually, after which DiskUsage will remain usable.

hsanjuan:
I think:

* It's ok to provide estimates as good as possible within a time frame of 3-5 minutes
* We already provide a method for the user to fix the estimates if they are willing. It consists in writing the `diskUsage.cache` file with the right value. Maybe we can make this more explicit with a special readme or printing a message.

Please let's remember how ***badly broken this is right now*** and take this as a step to improving this.

The vast majority of daemons with small storages attached won't be affected by the accuracy problems. Any new ipfs daemons won't be affected. Only nodes with huge existing storages will be affected. In this nodes it's impossible to run `repo stat` right now.

kevina:
> Only nodes with huge existing storages will be affected. In this nodes it's impossible to run repo stat right now.

Then it that case I think a better solution is to return an undefined value for the DiskUsage possible informing the user on how to rectify it for large repos rather than return an estimate that can be 5-10% off without informing the user of the inaccuracy.

hsanjuan:
> without informing the user of the inaccuracy.

I will inform the user

kevina:
I think it would be better to to keep track of if the disk usage is an estimate or an accurate count and inform the user of which.  I am okay if we delay this to another P.R. as long as it gets done sometime soonish.

Also please note, that my option is just that.  I will not block this P.R. if it is impeding important progress and will not object if someone dismisses my review.

hsanjuan:
I have implemented time deadline in the last commit (set a default of 5 minutes).

It will look like this when starting the datastore and no cache file exists:
```
root@biham /ipfs/ipfs_master/repo/blocks # ./test 
Calculating datastore size. This might take 5m0s at most and will happen only once
WARN: It took to long to calculate the datastore size
WARN: The total size (7589482442898) is an estimation. You can fix errors by
WARN: replacing the diskUsage.cache file with the right disk usage in bytes and
WARN: re-opening the datastore
7589482442898 <nil>

The actual reference size in biham is: 7536515831332 (so it's pretty close).
```

kevina:
> It will look like this when starting the datastore and no cache file exists:

It is possible to miss this warning.  This information should at least be stored in the datastore somehow.  Providing an interface to get to determine if the value stored is an approximation or an exact value can be a separate p.r.

kevina:
I'm sorry, but what is the standard policy in regard to an author merging there own pull request?  Especially when it is non-trivial.

Also see #31.

PR #29, opened 2018-02-27
gx publish 1.1.8

hsanjuan:


hsanjuan:
Note tests failing because bad go version. This is getting fixed in #27

PR #33, opened 2018-03-11
Store note on accuracy of initial disk usage calculation in diskUsage notes

### LINKED ISSUE #31: Store Some Indication of How Accurate the Initial DiskUsage() Calculation was
#27 landed without any provision for strong how accurate the initial DiskUsage() calculation was.  I would really like to see that stored in the datastore before this code get used in production (i.e used by go-ipfs).

kevina:
Closes #31

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ datastore is not open), with the right disk usage value in size. I.e., in the da
     3919232394        .
     $ echo -n "3919232394" > diskUsage.cache
 
+The accuracy of the initial disk usage calculation is stores in the file `diskUsage.notes`, this file is for reference only and is
+currently not used in any other way.
 
 ## Contribute
 
--- a/README.md
+++ b/README.md
@@ -60,8 +60,10 @@ datastore is not open), with the right disk usage value in size. I.e., in the da
     3919232394        .
     $ echo -n "3919232394" > diskUsage.cache
 
-The accuracy of the initial disk usage calculation is stores in the file `diskUsage.notes`, this file is for reference only and is
-currently not used in any other way.
+The accuracy of the initial disk usage calculation is stored in the file `diskUsage.notes`.  This file is currently for reference only and
+not used in any other way.  If the initial calculation was accurate the file will contain the value `initial-exact`.  If some of the
+directories have too many entries and the disk usage for that directory was estimated based on the first 2000 entries, the file will contain
+`initial-approximate`.  If the calculation took too long and timed out as indicated above, the file will contain `initial-timed-out`.
 
 ## Contribute
 
--- a/flatfs.go
+++ b/flatfs.go
@@ -35,6 +35,9 @@ var (
 	// DiskUsageFile is the name of the file to cache the size of the
 	// datastore in disk
 	DiskUsageFile = "diskUsage.cache"
+	// DiskUsageNodes stores notes about the accuracy of the size of
+	// the diskusage calculation
+	DiskUsageNotes = "diskUsage.notes"
 	// DiskUsageFilesAverage is the maximum number of files per folder
 	// to stat in order to calculate the size of the datastore.
 	// The size of the rest of the files in a folder will be assumed
@@ -54,6 +57,27 @@ const (
 	opDelete
 	opRename
 )
+
+type initAccuracy string
+
+const (
+	exactA    initAccuracy = "initial-exact"
+	approxA   initAccuracy = "initial-approximate"
+	timedoutA initAccuracy = "initial-timed-out"
+)
+
+func combineAccuracy(a, b initAccuracy) initAccuracy {
+	if a == timedoutA || b == timedoutA {
+		return timedoutA
+	}
+	if a == approxA || b == approxA {
+		return approxA
+	}
+	if a == exactA && b == exactA {
+		return exactA
+	}
+	return ""
+}
 
 var _ datastore.Datastore = (*Datastore)(nil)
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -60,6 +60,8 @@ datastore is not open), with the right disk usage value in size. I.e., in the da
     3919232394        .
     $ echo -n "3919232394" > diskUsage.cache
 
+The accuracy of the initial disk usage calculation is stores in the file `diskUsage.notes`, this file is for reference only and is
```
hsanjuan: `stored`

convert.go
```diff
@@ -11,9 +11,9 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-datastore/query"
-	"github.com/jbenet/go-os-rename"
+	"gx/ipfs/QmXRKBQA4wXP7xWbFiZsR1GP4HV6wMDQ1aWFxZZ4uBcPX9/go-datastore"
```
hsanjuan: should not leave rewritten

flatfs_test.go
```diff
@@ -660,7 +669,7 @@ func testDiskUsageEstimation(dirFunc mkShardFunc, t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-
+
```
hsanjuan: gofmt

flatfs_test.go
```diff
@@ -429,6 +429,15 @@ func testDiskUsage(dirFunc mkShardFunc, t *testing.T) {
 	}
 	t.Log("duPostDelete:", duDelete)
 
+	// Make sure DiskUsageNotes files exists and the correct value
+	d, err := ioutil.ReadFile(filepath.Join(temp, flatfs.DiskUsageNotes))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(d) != "initial-exact\n" {
```
hsanjuan: Use constant instead of literal?

flatfs_test.go
```diff
@@ -687,6 +696,15 @@ func testDiskUsageEstimation(dirFunc mkShardFunc, t *testing.T) {
 	if diff > maxDiff {
 		t.Fatal("expected a better estimation within 5%")
 	}
+
+	// Make sure DiskUsageNotes files exists and the correct value
+	d, err := ioutil.ReadFile(filepath.Join(temp, flatfs.DiskUsageNotes))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(d) != "initial-approximate\n" {
```
hsanjuan: Same

README.md
```diff
@@ -60,6 +60,8 @@ datastore is not open), with the right disk usage value in size. I.e., in the da
...
+currently not used in any other way.
```
hsanjuan: Let's explain what each of the three possible values mean, from a user perspective it's may not be clear what `initial-timed-out` actually signals.

kevina: Sure.

flatfs_test.go
kevina: It's a test, so I want to make sure it has the right value.

flatfs_test.go
kevina: It's a test, so I want to make sure it has the right value.

convert.go
kevina: Ooops.   Sorry.

flatfs_test.go
kevina: I thought I did.  Sorry.

flatfs_test.go
hsanjuan: But the right value is and will always be the one defined by the constant?

kevina: Yes, unless its accidentally modified in which case this test will fail to alert you of that fact.  If it was intentional, then the test case (and the README) should also be modified.

flatfs.go
```diff
@@ -35,6 +35,9 @@ var (
 	// DiskUsageFile is the name of the file to cache the size of the
 	// datastore in disk
 	DiskUsageFile = "diskUsage.cache"
+	// DiskUsageNodes stores notes about the accuracy of the size of
+	// the diskusage calculation
+	DiskUsageNotes = "diskUsage.notes"
```
Stebalien: Any reason not to put this in the same file? Or just have one JSON/CBOR file with all the information (significantly more flexible)? That would also allow it to be atomic.

This would also allow us to keep track of the disk size accuracy if we crash. We could record the fact that we're currently running *while* running. If we start up and see that we didn't correct the state, we can adjust the accuracy to approximate (or something else).

kevina: I did it this way for simplicity.  It has the advantage of not having to read the accuracy data once written, but I have no problem doing something more sophisticated.

flatfs.go
```diff
@@ -706,11 +736,11 @@ func (fs *Datastore) calculateDiskUsage() error {
 
 	fmt.Printf("Calculating datastore size. This might take %s at most and will happen only once\n", DiskUsageCalcTimeout.String())
 	deadline := time.Now().Add(DiskUsageCalcTimeout)
-	du, err := folderSize(fs.path, deadline)
+	du, accuracy, err := folderSize(fs.path, deadline)
 	if err != nil {
 		return err
 	}
-	if time.Now().After(deadline) {
+	if accuracy == timedoutA {
```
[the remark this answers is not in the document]
kevina: Maybe.  My code does not change the behavior of @hsanjuan code.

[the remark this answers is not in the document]
Stebalien: Fair enough. We can always change this later.

flatfs.go
Stebalien: My motivation is really atomicity. It allows us to atomically store the repo size *and* it's accuracy (and allows us to deal with the inaccurate repo size on crash case). However, I don't really feel strongly any which way.

kevina:
Done, please see #35.

kevina:
Closing in favor of #35.

PR #34, opened 2018-03-15
Change go-os-rename to standard

djdv:
Since Go 1.5, os.Rename on Windows is consistent with other platforms. `go-os-rename` just enforced consistent behaviour on Windows before the change was made upstream, for other platforms it used the standard. Since both are the same now it seems redundant to keep it.


Kubuxu:
I wasn't 100% sure that is why I waited.

PR #35, opened 2018-03-23
Use JSON file for `diskUsage.cache` and include note on accuracy of value

kevina:
Alternative to #33.


flatfs.go
```diff
@@ -79,14 +100,24 @@ type Datastore struct {
 	// sychronize all writes and directory changes for added safety
 	sync bool
 
-	diskUsage           int64
-	diskUsageCheckpoint int64
+	diskUsage   int64
+	storedValue diskUsageValue
+	// updateLock must be held when updating storedValue or writing
+	// to a file, it doesn't need to be held when reading
```
Stebalien: to "a file"?

flatfs.go
```diff
@@ -756,43 +797,54 @@ func (fs *Datastore) checkpointDiskUsage(newDuInt int64) {
...
+	origVal := fs.storedValue.diskUsage
+	// update the stored diskUsage value now to prevent unnecessary
+	// calls to persistDiskUsageFile.  On error role back the value to
+	// the original
+	atomic.StoreInt64(&fs.storedValue.diskUsage, du)
 
-	duB := []byte(fmt.Sprintf("%d", du))
 	tmp, err := ioutil.TempFile(fs.path, "du-")
 	if err != nil {
+		atomic.StoreInt64(&fs.storedValue.diskUsage, origVal)
```
Stebalien: Should definitely be logging these errros.

flatfs.go
```diff
@@ -756,43 +797,54 @@ func (fs *Datastore) checkpointDiskUsage(newDuInt int64) {
...
+// persistDiskUsageFile updates the diskusage file with the last known
+// value
 func (fs *Datastore) persistDiskUsageFile() {
-	// Store DiskUsage on clean shutdowns.
-	du, err := fs.DiskUsage()
-	if err != nil {
-		// do not store on errors. just ignore them
-		return
-	}
+	fs.updateLock.Lock()
```
Stebalien: You know, really, we should probably just do this in a goroutine and signal the goroutine with a channel. We could end up with a lot of blocking here, especially because this the "should we update" check is racy. (old bug, I guess, but we should still fix it).

flatfs.go
```diff
@@ -691,33 +728,37 @@ func folderSize(path string, deadline time.Time) (int64, error) {
...
 }
 
 // calculateDiskUsage tries to read the DiskUsageFile for a cached
 // diskUsage value, otherwise walks the datastore files.
+// it is only safe to call in Open()
 func (fs *Datastore) calculateDiskUsage() error {
 	// Try to obtain a previously stored value from disk
 	if persDu := fs.readDiskUsageFile(); persDu > 0 {
-		atomic.StoreInt64(&fs.diskUsage, persDu)
+		fs.diskUsage = persDu
```
Stebalien: Eventually, I'd like to *also* mark the on-file disk usage as an estimate until we close. That way, if we crash, we know that our estimate may be off by a bit. However, we can probably do that in a future PR (your preference).

flatfs.go
kevina: I mean when writing `diskUsage` to a file.

flatfs.go
kevina: Okay.

flatfs.go
kevina: Your right, and that was going to be my next step.

Kubuxu:
Won't this need migration?

Stebalien:
> Won't this need migration?

We haven't bubbled the previous changes yet.

flatfs.go
```diff
@@ -739,60 +790,100 @@ func (fs *Datastore) updateDiskUsage(path string, add bool) {
...
+				fs.done <- true
+				return
+			}
+			// If the difference between the checkpointed disk usage and
+			// current one is larger than than `diskUsageCheckpointPercent`
+			// of the checkpointed: store it.
+			newDu := float64(du)
+			lastCheckpointDu := float64(fs.storedValue.diskUsage)
+			diff := math.Abs(newDu - lastCheckpointDu)
+			if (lastCheckpointDu * diskUsageCheckpointPercent / 100.0) < diff {
```
Stebalien: ultra nit: can avoid floats here (lastCheckpointDu * diskUsageCheckpointPercent) < diff * 100

flatfs.go
```diff
@@ -739,60 +790,100 @@ func (fs *Datastore) updateDiskUsage(path string, add bool) {
...
+				fs.writeDiskUsageFile(du)
+			}
+			// Otherwise insure the value will be written to disk after
+			// `diskUsageCheckpointTimeout`
+			if fs.dirty && !timerActive {
+				timer.Reset(diskUsageCheckpointTimeout)
```
Stebalien: We don't appear to ever turn this off.

kevina: Oops.  Forgot to add `timerActive = true`.

flatfs.go
```diff
@@ -739,60 +789,82 @@ func (fs *Datastore) updateDiskUsage(path string, add bool) {
...
 		return
 	}
 
-	os.Rename(tmp.Name(), filepath.Join(fs.path, DiskUsageFile))
+	if err := os.Rename(tmp.Name(), filepath.Join(fs.path, DiskUsageFile)); err != nil {
+		return
+	}
+
+	fs.storedValue.diskUsage = du
+	fs.dirty = false
```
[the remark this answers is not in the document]
kevina: I tried that, and it seamed to make the code more complicated.

flatfs.go
kevina: We can avoid the division but not the use of floats.  `diskUsageCheckpointPercent` is a float and I don't want to make it an integer.

flatfs.go
Stebalien: Still missing a call to Stop as far as I can tell (and a defered call to Stop).

kevina: As far as I can tell, there is no need to call Stop here.  If timerActive is false than the timer already expired.  From the go docs (https://golang.org/pkg/time/#Timer.Reset):
> If a program has already received a value from t.C, the timer is known to have expired, and t.Reset can be used directly.

kevina:
Note. I enhanced the tests and am now getting this failure:
```
--- FAIL: TestDiskUsage (0.95s)
    --- FAIL: TestDiskUsage/next-to-last (0.32s)
        flatfs_test.go:401: /tmp/test-datastore-flatfs-557319926
        flatfs_test.go:416: duNew: 1238
        flatfs_test.go:433: duPostPut: 4038
        flatfs_test.go:448: duPostDelete: 2038
        flatfs_test.go:451: duFinal: 2038
        flatfs_test.go:470: Unexpected value for diskUsage in diskUsage.cache: 2058 (expected 2038)
```
I put debug statements in the code and discovered that the value stored in `diskUsage.cache` is a value from a previous write.  I verified that the final write to disk on close is being done in the correct order and has the expected value.  Right now my only conclusion is that the o.s. is doing the renaming from the temporary files out of order and an older rename is being done before the final, but I can't seam to find any documentation that this is ever a problem is the write to temp file and rename over the original.  I do know that you should probably sync the temporary file before close but I tried that and it didn't seam to help.

I will look more into this tomorrow, but suggestions welcome. Do I need to sync the directory also?

kevina:
> I think many checkpoints are discarded because the channel is full (// checkpoint request already pending). Thus you may be discarding the last one with the expected size in the test because previous ones are pending.
That is the intended behavior, it keeps unnecessary ones from piling up. The final checkpoint is not discarded because it only happens when the channel is closed.  I verified this with debug print statements that I didn't push.

kevina:
I found the bug, I was doing something stupid.  This should be ready for review now.

hsanjuan:
Are we sure this is not overkill in terms of code and complexity just to inform the user that the disk usage is approximate (which always is anyway) ?

I can imagine that either users care about accuracy, in which case, given we cannot tell them how far the approximation is from reality, they'll want to manually calculate and fix their diskUsage files. Or don't care so much, in which case it is the same how the approximation was performed. This might be solving a pure documentation issue with complex code and I'm not sure it's the best thing. Just my two cents...

kevina:
@hsanjuan this was mostly @Stebalien idea, it also solves some other problems such as doing the checkpoint asynchronously and ensuring the disk usage is written to disk every 2 seconds.

Stebalien:
@hsanjuan basically, the only *complex* changes here are performance related. The "use a single json file" change should actually simplify things (although my motivation *was* to allow us to atomically update the accuracy and the size).

flatfs.go
```diff
@@ -79,14 +101,27 @@ type Datastore struct {
 	// sychronize all writes and directory changes for added safety
 	sync bool
 
-	diskUsage           int64
-	diskUsageCheckpoint int64
+	// atmoic operations should always be used with diskUsage
+	diskUsage int64
```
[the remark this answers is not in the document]
Stebalien: Hm. That's something that should go a the *top* of the documentation and in big bold letters everywhere else.

flatfs.go
```diff
@@ -54,6 +55,27 @@ const (
...
+		return timedoutA
+	}
+	if a == approxA || b == approxA {
+		return approxA
+	}
+	if a == exactA && b == exactA {
+		return exactA
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
kevina: There shouldn't be any other cases.  The best I can do is if one is an empty string return the other.  If a and b disagree otherwise the is no good value to return so I will just return "".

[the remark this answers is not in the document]
Stebalien: Then I'd return some form of "unknown".

flatfs.go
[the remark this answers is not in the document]
whyrusleeping: > Hm. That's something that should go a the top of the documentation and in big bold letters everywhere else.

Thats what *I* said. Its actually at the bottom, in normal font:
https://golang.org/pkg/sync/atomic/#pkg-note-BUG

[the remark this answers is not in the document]
Kubuxu: @kevina no it depends on how Go lays out the structure and x86-32 does not have atomic ops on misaligned 64bit variables.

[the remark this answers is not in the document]
kevina: @Kubuxu I think we are saying the same thing.  See http://www.tapirgames.com/blog/golang-memory-alignment

flatfs.go
```diff
@@ -4,30 +4,32 @@
...
 	"path/filepath"
-	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
-	"github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-datastore/query"
+	"gx/ipfs/QmeiCcJfDW1GJnWUArudsv5rQsihpi4oyddPhdqo3CfX6i/go-datastore"
```
Stebalien: Shouldn't be rewritten.

flatfs.go
Stebalien: Ah. I see.

flatfs.go
```diff
@@ -739,60 +810,110 @@ func (fs *Datastore) updateDiskUsage(path string, add bool) {
...
+		}
+	}
+
 	if err := tmp.Close(); err != nil {
+		log.Warningf("cound not write disk usage: %v", err)
+		return
+	}
+
+	if err := os.Rename(tmp.Name(), filepath.Join(fs.path, DiskUsageFile)); err != nil {
+		log.Warningf("cound not write disk usage: %v", err)
```
Stebalien: This should probably be deleting the temporary file on failure (or at least trying to do so).

flatfs.go
kevina: Oops! I have been using `git gui` and hand picking the chunks to include, looks like I wasn't careful enough.

flatfs.go
kevina: Done.

PR #38, opened 2018-03-27
Create Jenkinsfile

victorb:


VictorBjelkholm:
Windows error:

```
Calculating datastore size. This might take 5m0s at most and will happen only once
Calculating datastore size. This might take 5m0s at most and will happen only once
convert_test.go:92: Could not undo the move.%!(EXTRA *errors.errorString=C:\Windows\TEMP\test-datastore-flatfs-047642534\v1: SHARDING file not found in datastore)
flatfs_test.go:32: tempdir cleanup failed: remove C:\Windows\TEMP\test-datastore-flatfs-047642534\v1\xa5: Access is denied.
```

djdv:
@VictorBjelkholm 
`C:\Windows\TEMP` is not usually writable by regular users, ioutil.TempDir() will check these environment variables in order: `TMP`, `TEMP`, `USERPROFILE ` and finally default to `C:\Windows\TEMP`
These are set for a user and not typically the system.

The test environment should probably have `TMP` set to a writable directory. I beleive this is the default value that is set for users `%LOCALAPPDATA%\Temp` which in cmd would be `set TMP %LOCALAPPDATA%\Temp`, but it could be set to anywhere desirable for temporary data.

PR #47, opened 2018-09-10
check error before defer-removing disk usage file

b5:
Ran into this while updating our gx deps:
https://circleci.com/gh/qri-io/cafs/43

Not being able to write the temp file in the first place is the source of the failure, but handling errors before invoking defer statements on nil pointers is a good idea IMHO 😉

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -863,6 +863,11 @@ func (fs *Datastore) checkpointLoop() {
 
 func (fs *Datastore) writeDiskUsageFile(du int64, doSync bool) {
 	tmp, err := ioutil.TempFile(fs.path, "du-")
+	if err != nil {
+		log.Warningf("cound not write disk usage: %v", err)
+		return
+	}
+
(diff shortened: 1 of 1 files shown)


flatfs.go
```diff
@@ -864,16 +864,17 @@ func (fs *Datastore) checkpointLoop() {
 func (fs *Datastore) writeDiskUsageFile(du int64, doSync bool) {
 	tmp, err := ioutil.TempFile(fs.path, "du-")
 	removed := false
+	if err != nil {
+		log.Warningf("cound not write disk usage: %v", err)
+		return
+	}
+
```
kevina: Nit: This should come even earlier, before imminently after the call, and before `removed := false`

b5:
👋 @kevina! Hope all is well in your neck of the woods 😄. Just force-pushed your requested changes.

kevina:
Thanks.

Merged.

PR #52, opened 2019-02-28
fix panic on write after close

Stebalien:
(and make Close threadsafe)
(and fix some error handling)

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -476,7 +476,7 @@ func (fs *Datastore) doPut(key datastore.Key, val []byte) error {
 	return nil
 }
 
-func (fs *Datastore) putMany(data map[datastore.Key]interface{}) error {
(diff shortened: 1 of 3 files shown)


flatfs.go
```diff
@@ -465,15 +476,32 @@ func (fs *Datastore) doPut(key datastore.Key, val []byte) error {
...
-			return datastore.ErrInvalidType
+	files := make(map[*os.File]*op, len(data))
+	ops := make(map[*os.File]int, len(data))
+
+	defer func() {
+		for fi := range files {
+			val, _ := ops[fi]
+			switch val {
+			case 0:
+				_ = fi.Close()
```
magik6k: It would be great if we figured out a way to do this in smaller chunks as every batch put will use as much FDs as there are entries. Out of scope for this PR though.

PR #60, opened 2019-12-03
Update Datastore Interface

aschmahmann:
Per ipfs/go-datastore#140 the Datastore interface is being updated to support asynchronous writes.

This PR should be approved before the above PR is merged. After ipfs/go-datastore#140 is merged then this PR should have its go.mod updated and merged.


flatfs.go
```diff
@@ -388,6 +388,16 @@ func (fs *Datastore) Put(key datastore.Key, value []byte) error {
...
 }
 
+func (fs *Datastore) Sync(prefix datastore.Key) error {
+	fs.shutdownLock.RLock()
+	defer fs.shutdownLock.RUnlock()
+	if fs.shutdown {
+		return ErrClosed
+	}
+
+	return nil
```
aschmahmann: Is this the approach we want here? I see that FlatFS has support for both sync and async operation (async seems to be the default, but IPFS uses the sync version by default).

Do we want to support some sort of sync here in the event that the datastore is running in async mode? If so is it important enough to do something now or should I just file an issue?

Stebalien: Let's file an issue and leave this as-is for now.

 A correct implementation would need to call fsync on all changed files. Doing so would require keeping track of file changes (or relying on timestamps).

PR #32, opened 2018-03-09
Add a proper README.md for this project.

hsanjuan:
Include diskUsage notes.

Issue #61, opened 2019-12-05
Document hierarchical keys bug in README


aschmahmann:
It doesn't look like #22 is going to be fixed in the immediate future, so we should at least make it clear in the README that Flatfs does not support hierarchical keys.

aschmahmann:
Looks like this was closed by @Stebalien in #64

Issue #40, opened 2018-05-08
User-friendly keys


Stebalien:
Two proposals (assuming we don't deprecate this datastore soon):

1. Handle base32 encoding internally. Forcing the client to deal with filesystem name restrictions is painful and unnecessary.
2. Shard using a non-cryptographic hash function. We could, e.g., use SipHash (at most 50ns per hash versus 4000ns to open a file). Again, we shouldn't be making assumptions about keys specified by the user.

This way, we'd map `key` to `Base32(FastHash(key))[:2] + "/" + Base32(key)`.

kevina:
We originally used binary keys and encoded to base16 in flatfs and had binary keys, that created some serious problems, see https://github.com/ipfs/go-ipfs/issues/2601 for context.

Stebalien:
Well, yes, we shouldn't be mangling keys. I'd just treat them as byte strings.

kevina:
The problem was that we use '/' as a path separator in datastore keys, that does not mix well when the key itself in binary, so we changed from using binary keys to text and encoded them to base32 before converting them to datastore keys.

Stebalien:
Not sure if I understand. The `/` path separator used in *keys* shouldn't be reflected on-disk.

aschmahmann:
Looks like this was closed by @Stebalien in #64

PR #64, opened 2020-02-11
Make flatfs robust

Stebalien:
1. Improve query logic to finally support all features. Offsets, limits, filters, and sorting are all supported through naive queries.
2. Fetch files from disk when asked. Not exactly fast but the user is allowed to dig their own grave.
3. Explicitly restrict keys instead of randomly doing bad things at runtime.

Importantly, this change only allows keys of the form `/[0-9A-Z+-_=]`. That is, upper-case alphanumeric keys in the root namespace (plus some special characters).

Why? We don't encode keys before writing them to the filesystem. This change ensures that:

1. Case sensitivity doesn't matter because we only allow upper-case keys. 
2. Path separators and special characters doesn't matter.

For context, go-ipfs only uses flatfs for storing blocks. Every block CID is encoded as uppercase alphanumeric text (specifically, uppercase base32).

We could be less restrictive, but this is safer and easier to understand. Unfortunately, we _can't_ allow mixed case (Windows) and can't allow lowercase because we're already using uppercase keys.

---

fixes #32, #61, #40, #22

tests with the query test suite


flatfs.go
```diff
@@ -1042,11 +1071,30 @@ func (fs *Datastore) walk(path string, result *query.ResultBuilder) error {
...
 		}
 
+		var (
+			value []byte
+			size  int
+			err   error
+		)
+		if !result.Query.KeysOnly {
+			value, err = ioutil.ReadFile(filepath.Join(path, fn))
+			size = len(value)
```
dirkmc: Is it safe to assume that if there's an error, value will have length 0? (seems like it should be, just want to confirm)

key_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"testing"
+
+	"github.com/ipfs/go-datastore"
+)
+
+var (
+	validKeys = []string{
+		"/FOO",
+		"/1BAR1",
+		"/=EMACS-IS-KING=",
```
hsanjuan: :D

flatfs.go
```diff
@@ -1090,12 +1138,17 @@ func (fs *Datastore) Batch() (datastore.Batch, error) {
 }
 
 func (bt *flatfsBatch) Put(key datastore.Key, val []byte) error {
+	if !keyIsValid(key) {
+		return fmt.Errorf("key not supported by flatfs: '%q'", key)
```
hsanjuan: Nitpick: Might add a new `ErrInvalidKey` to the top and wrap it.

flatfs.go
```diff
@@ -682,7 +700,18 @@ func (fs *Datastore) Query(q query.Query) (query.Results, error) {
...
+	results := b.Results()
+
+	if len(q.Filters) > 0 ||
+		len(q.Orders) > 0 ||
+		q.Limit > 0 ||
+		q.Offset > 0 {
+		nq := q
+		nq.Prefix = "" // already handled
+		results = query.NaiveQueryApply(nq, results)
+	}
```
hsanjuan: So `NaiveQueryApply` will already do nothing if it does not have to do anything so you might skip the checks here as redundant: https://github.com/ipfs/go-datastore/blob/master/query/query_impl.go#L119

flatfs.go
```diff
@@ -654,17 +678,11 @@ func (fs *Datastore) doDelete(key datastore.Key) error {
...
-		!q.KeysOnly ||
-		q.ReturnExpirations ||
-		q.ReturnsSizes {
-		// TODO this is overly simplistic, but the only caller is
-		// `ipfs refs local` for now, and this gets us moving.
-		return nil, errors.New("flatfs only supports listing all keys in random order")
+	if prefix != "/" {
+		// This datastore can't include keys with multiple components.
+		// Therefore, it's always correct to return an empty result when
+		// the user requests a filter by prefix.
```
hsanjuan: Log a warning message? Very likely indicates bad usage of the datastore which might happen silently to the application above.

key_test.go
Stebalien: Clearly.

flatfs.go
Stebalien: Hm. Good question. No. It'll return the partial value and an error.  I'll change this.

flatfs.go
Stebalien: Ah, well, https://github.com/ipfs/go-datastore/pull/147 (now merged).

hsanjuan: You are setting prefix to `""` anyways so it would have not entered the `if`

Stebalien:
> Any relevant context/request that makes this a necessary change or is it just general maintenance?

General maintenance. I needed to pull through a fix for prefix handling in the query logic.

hsanjuan:
Errors in tests, seems circle is building with old Go ?

Stebalien:
> Errors in tests, seems circle is building with old Go ?

Updated go-log.

Stebalien:
Oh, and probably old go.

Issue #70, opened 2020-04-09
File mode on windows causes issues between renaming and writes


willscott:
ref: https://github.com/ipfs/go-ipfs/issues/7115#event-3219629851

Current attempt towards mitigation is to open files on windows with the SHARE_DELETE flag per https://github.com/golang/go/issues/32088

willscott:
After the change in #71, the following windows errors occur in the test suite:
```
--- FAIL: TestMoveRestart (0.96s)
    convert_test.go:92: Could not undo the move. C:\Users\circleci\AppData\Local\Temp\test-datastore-flatfs-525984286\v1: SHARDING file not found in datastore
    flatfs_test.go:34: tempdir cleanup failed: remove C:\Users\circleci\AppData\Local\Temp\test-datastore-flatfs-525984286\v1\XTV: Access is denied.
--- FAIL: TestSuite (84.81s)
    --- FAIL: TestSuite/github.com/ipfs/go-datastore/test.SubtestCombinations (77.27s)
        --- FAIL: TestSuite/github.com/ipfs/go-datastore/test.SubtestCombinations/1/{SELECT·keys,vals·FILTER·[KEY·<·"/2"]·ORDER·[KEY]} (0.00s)
            basic_tests.go:430: putting 4 values
            basic_tests.go:438: getting values back
            basic_tests.go:450: querying values
            basic_tests.go:460: aggregating query results
            basic_tests.go:463: query result error:  walk failed: open C:\Users\circleci\AppData\Local\Temp\test-datastore-flatfs-599044958\diskUsage.cache: The process cannot access the file because it is being used by another process.
        --- FAIL: TestSuite/github.com/ipfs/go-datastore/test.SubtestCombinations/1/{SELECT·keys,vals·FILTER·[KEY·<·"/2"]·ORDER·[KEY]·OFFSET·95·LIMIT·1} (0.00s)
            basic_tests.go:430: putting 4 values
            basic_tests.go:438: getting values back
            basic_tests.go:450: querying values
            basic_tests.go:460: aggregating query results
            basic_tests.go:463: query result error:  walk failed: open C:\Users\circleci\AppData\Local\Temp\test-datastore-flatfs-599044958\diskUsage.cache: The process cannot access the file because it is being used by another process.
[...]

PR #71, opened 2020-04-09
open files on windows with shared delete mode

willscott:
address #70 

suggestions welcome on any hints on the set of operations into the datastore that would trigger a windows failure before but we are hopeful might not anymore

### DIFF
--- a/util_windows.go
+++ b/util_windows.go
@@ -1,0 +1,93 @@
+// +build windows
+
+package flatfs
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	goissue34681 "github.com/alexbrainman/goissue34681"
+)
+
+var tmpRand uint32
+var randmu sync.Mutex
+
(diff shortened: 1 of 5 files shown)


util_windows.go
```diff
@@ -0,0 +1,93 @@
+// +build windows
```
Stebalien, line 1: let's document that we got this from golang (license, attribution, etc.)

util_windows.go
```diff
@@ -0,0 +1,93 @@
...
+func prefixAndSuffix(pattern string) (prefix, suffix string) {
+	if pos := strings.LastIndex(pattern, "*"); pos != -1 {
+		prefix, suffix = pattern[:pos], pattern[pos+1:]
+	} else {
+		prefix = pattern
+	}
+	return
+}
+
+func tempFile(dir, pattern string) (f *os.File, err error) {
```
Stebalien, line 46: Actually, now that we're doing this, we should rename the temporary file _while it's open_. If we do that, nobody will be able to re-open the file _without_ the share delete flag.

willscott:
subsumed by #73

PR #74, opened 2020-04-10
fix: close query when finished moving

Stebalien:


convert.go
```diff
@@ -76,11 +76,13 @@ func Move(oldPath string, newPath string, out io.Writer) error {
 			break
 		}
 		if e.Error != nil {
+			res.Close()
```
willscott, line 79: could you defer a `.Close()` instead?

Stebalien: I want to close this _before_ we finish the move, and I don't want to close it twice. I could have wrapped this in a function, but eh...

PR #75, opened 2020-04-10
add fd-overload retries inline

willscott:
pull in from the higher level ds-datastore retry wrapper, so that retries can happen appropriately within the putMany operation.

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -586,8 +586,10 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
 
 		// Fallback retry for temporary error.
 		if err != nil && isTooManyFDError(err) {
+			if err = closer(); err != nil {
+				return err
+			}
 			for i := 0; i < 6; i++ {
-				closer()
 				time.Sleep(time.Duration(i+1) * RetryDelay)
 
 				tmp, err = fs.tempFile()
@@ -614,7 +616,10 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
 
 	// Now we sync everything
 	// sync and close files
-	closer()
+	err := closer()
+	if err != nil {
+		return err
+	}
 
 	// move files to their proper places
 	for fi, op := range files {
--- a/flatfs.go
+++ b/flatfs.go
@@ -460,18 +460,14 @@ func (fs *Datastore) doWriteOp(oper *op) error {
 	}
 
 	// Do the operation
-	err := fs.doOp(oper)
+	var err error
+	for i := 0; i < 6; i++ {
+		err = fs.doOp(oper)
 
-	// Fallback retry for temporary error.
-	if err != nil && isTooManyFDError(err) {
-		for i := 0; i < 6; i++ {
-			time.Sleep(time.Duration(i+1) * RetryDelay)
-
-			err = fs.doOp(oper)
-			if err == nil || !isTooManyFDError(err) {
-				break
-			}
+		if err == nil || !isTooManyFDError(err) {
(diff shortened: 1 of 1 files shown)


flatfs.go
```diff
@@ -540,6 +565,18 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
 		dirsToSync = append(dirsToSync, dir)
 
 		tmp, err := fs.tempFile()
+
+		// Fallback retry for temporary error.
+		if err != nil && isTooManyFDError(err) {
```
Stebalien, line 570: I think we actually need to finish putting the _open_ files here. Otherwise, if there are multiple callers, they'll just all wait on each other to finish the put, then all bail.

Anyways, we should probably have some kind of "I have over 100 files open, time to write them to disk" logic.

flatfs.go
```diff
@@ -540,6 +565,18 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
...
+			for i := 0; i < 6; i++ {
+				time.Sleep(time.Duration(i+1) * RetryDelay)
+
+				tmp, err = fs.tempFile()
+				if err == nil || !isTooManyFDError(err) {
+					break
+				}
+			}
+		}
```
Stebalien, line 579: Nit: there's a lot of repeated code here.

flatfs.go
```diff
@@ -449,6 +462,18 @@ func (fs *Datastore) doWriteOp(oper *op) error {
...
+	if err != nil && isTooManyFDError(err) {
+		for i := 0; i < 6; i++ {
+			time.Sleep(time.Duration(i+1) * RetryDelay)
+
+			err = fs.doOp(oper)
+			if err == nil || !isTooManyFDError(err) {
+				break
+			}
+		}
+	}
```
Stebalien, line 475: nit: there's a lot of repeated code here. We can do the first put attempt _inside_ the loop.

flatfs.go
```diff
@@ -601,6 +638,24 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
...
+	if err != nil && isTooManyFDError(err) {
+		for i := 0; i < 6; i++ {
+			time.Sleep(time.Duration(i+1) * RetryDelay)
+
+			value, err = fs.get(key)
+			if err == nil || !isTooManyFDError(err) {
+				break
+			}
+		}
+	}
```
Stebalien, line 653: ditto repeated code.

flatfs.go
```diff
@@ -601,6 +638,24 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
 }
 
 func (fs *Datastore) Get(key datastore.Key) (value []byte, err error) {
+	value, err = fs.get(key)
```
Stebalien, line 641: Hm. Maybe push the retry loop all the way down into `readFile`? That could save us some work.

flatfs.go
Stebalien: We can do this in this patch, or a new patch. Up to you.

flatfs.go
```diff
@@ -540,6 +583,19 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
 		dirsToSync = append(dirsToSync, dir)
 
 		tmp, err := fs.tempFile()
+
+		// Fallback retry for temporary error.
+		if err != nil && isTooManyFDError(err) {
+			for i := 0; i < 6; i++ {
+				closer()
```
Stebalien, line 590: Only needs to be called once.

PR #78, opened 2020-04-10
feat: read harder

Stebalien:
Instead of retrying in get only, retry everywhere we read a file.


Stebalien:
This will fix other places where we read file. E.g., if we query the datastore.

PR #79, opened 2020-04-10
Make puts "succeed" if they wouldn't change the file data anyways

Stebalien:
This works around a windows issue where some other process sees the new file and decides to open it in the wrong sharing mode.

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -4,6 +4,7 @@
 package flatfs
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


flatfs.go
```diff
@@ -365,6 +366,29 @@ func (fs *Datastore) renameAndUpdateDiskUsage(tmpPath, path string) error {
...
+	// We failed to rename. In many cases, we're dealing content addressed
+	// data (blocks) so, if there's some weird filesystem error where we
+	// failed a rename (e.g., some other windows process is reading the
+	// file...), just check to see if we _would_ have done anything. If not,
+	// walk away.
+	//
+	// Is this slow? Yes. But it's not common so... whatever. Unfortunately,
+	// it _is_ common enough for windows users to cause problems if we
+	// _don't_ have this check.
+	if err != nil && fi != nil {
```
willscott, line 379: can we limit to a specific error?

Stebalien:
I'd like to do the dumb thing for now. If we notice this showing up in all those windows performance profiles we'll never actually collect, then we can try to make it better. But I'm fine if the failure path is a bit slower in this case.

Stebalien:
We're going to just repeatedly retry the put.

PR #80, opened 2020-04-13
cleanup putMany implementation

Stebalien:
* Use a map for dirsToSync to avoid syncing the same dir multiple times.
* Keep track of files in a slice, and use offsets into the slice to keep track
of which ones have been closed/removed.

Also, record the fact that we've created a temporary file _before_ we try to write to it, in case the write fails. That way, we'll try to remove it when we abort.

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -537,42 +537,43 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
 		return ErrClosed
 	}
 
-	var dirsToSync []string
+	type putManyOp struct {
+		key     datastore.Key
+		file    *os.File
+		dstPath string
+		srcPath string
+	}
 
-	files := make(map[*os.File]*op, len(data))
-	ops := make(map[*os.File]int, len(data))
+	var (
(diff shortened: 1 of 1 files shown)


flatfs.go
```diff
@@ -624,19 +624,24 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
...
+			tmp:  pop.file.Name(),
+			path: pop.path,
+		})
 		if err != nil {
 			return err
-		} else if done {
-			// signify removed
-			ops[fi] = 2
+		} else if !done {
+			_ = os.Remove(pop.file.Name())
```
willscott, line 637: I've convinced myself this works, but there are two things suboptimal we might want to make better:
* The distinction between `pop.path` and `pop.file.Name()` is not well named - which is previous vs new location?
* We're accessing a method (`.Name()`) after `.Close()`ing files. That's relying on the internal behavior to not clear all state when closing the handle, which an implementation would be within its rights to do.

Stebalien:
@willscott good point. I've addressed your feedback.

PR #81, opened 2020-04-17
alway take backoff retry loop on rename operations

willscott:
there may be some types of errors (i'm thinking of "ErrNotExist") that would be worth short-circuiting on.


Stebalien:
Looking through this, I believe we've been doing the retries at the wrong level: https://github.com/ipfs/go-ds-flatfs/pull/82

willscott:
closing in favor of #82

PR #82, opened 2020-04-17
move retries lower and retry rename ops

Stebalien:


Stebalien:
@willscott could you check this logic thoroughly. I think I got everything right, but I'm not sure.

flatfs.go
```diff
@@ -364,13 +368,21 @@ func (fs *Datastore) renameAndUpdateDiskUsage(tmpPath, path string) error {
 	// Rename and add new file's diskUsage. If the rename fails,
 	// it will either a) Re-add the size of an existing file, which
 	// was sustracted before b) Add 0 if there is no existing file.
-	err = os.Rename(tmpPath, path)
+	for i := 0; i < RetryAttempts; i++ {
+		err = os.Rename(tmpPath, path)
+		// if there's no error, or the source file doesn't exist, abort.
+		if err == nil || os.IsNotExist(err) {
```
willscott, line 374: Not sure if it's worth going down this rabbit hole of understanding what classes of errors are possible here.
* it looks like you'll get an `EEXIST` error in some cases if you try renaming a file to itself.
* renaming across disks (https://stackoverflow.com/questions/50740902/move-a-file-to-a-different-drive-with-go) can fail in some cases. I wonder what happens when your tmp dir is on a different disk than your main data cache.

Stebalien:
> We were also retrying put/delete write operations, which are no longer retried. I think transient failures in those aren't as much as much of a problem, but noting the functional change.

Yeah. I don't think we should be retrying write operations, but we probably should retry remove operations.

flatfs.go
```diff
@@ -364,13 +368,21 @@ func (fs *Datastore) renameAndUpdateDiskUsage(tmpPath, path string) error {
...
+			break
+		}
+		// Otherwise, this could be a transient error due to some other
+		// process holding open one of the files. Wait a bit and then
+		// retry.
+		time.Sleep(time.Duration(i+1) * RetryDelay)
+	}
 	fs.updateDiskUsage(path, true)
```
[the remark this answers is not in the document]
Stebalien: I don't know exactly why we do this, but I believe it's because we're concerned that the rename may not be atomic. That is, we:

1. Decrement the disk usage.
2. Rename.
3. Fix the disk usage _always_.

This will work even if the rename returns an error, but actually updates the target file under the covers.

flatfs.go
Stebalien: We're putting the temporary directory _inside_ the datastore directory, so that shouldn't be an issue.

Really, we probably don't even need "is not exist", but we might as well have it.

Stebalien:
Ok, I've added retries to the delete ops as well.

PR #83, opened 2020-04-24
Add NewDatastore entry point

willscott:
the fuzzer defaults to instantiating datastore implementations using `NewDatastore`, and interface method present in most of the other datastore implementations. Proposing adding here for ease of fuzzing.

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -297,6 +297,11 @@ func CreateOrOpen(path string, fun *ShardIdV1, sync bool) (*Datastore, error) {
 	return Open(path, sync)
 }
 
+// NewDatastore is a common interface to go-datastore for opening with default options.
+func NewDatastore(path string, options *interface{}) (*Datastore, error) {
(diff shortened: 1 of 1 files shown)


Stebalien:
Is there any way to avoid adding this to the public interface?

willscott:
a manual stub for flatfs can be added in `go-datastore/fuzz`. It seems like since this repo a convention has emerged of naming the entry point `NewDatastore`, with functional store-specific options, versus the implementation specific entry point used by flatfs currently.

Stebalien:
I don't think we're going to be able to define a good abstraction here:

* `*interface{}` is not a friendly abstraction for options.
* Not all datastores will necessarily have paths.

willscott:
*interface{} was a placeholder for whatever option type we want - it could just as easily be a `*ShardIdV1` since that's the current option taken by flatfs. The point is to have it structured such that a `nil` parameter provides sane defaults, as in badger / leveldb / etc.

Point taken though, and I'll do this as a stub in the fuzz code.

PR #87, opened 2020-07-27
Add os.Rename wrapper for Plan 9

### LINKED ISSUE #86: Tests are failing on Plan 9 because os.Rename doesn't rename across directory
Example failure:
```
cpu% go test -run 'TestMove$'
...
--- FAIL: TestMove (4.24s)
    convert_test.go:216: Put fail: rename /tmp/test-datastore-flatfs-825411366/v1/.temp/temp-030917448 /tmp/test-datastore-flatfs-825411366/v1/XKC/XKCILQAO5ZLD2K===.data: invalid argument
```

fhs:
os.Rename documentation says: "OS-specific restrictions may apply when
oldpath and newpath are in different directories." On Unix, this means
we can't rename across devices. On Plan 9 however, the functionality is
even more limited: cross-directory renames are not allowed at all.

Add a wrapper around os.Rename for Plan 9, which will copy the file if
we're renaming across directory. All tests seems to pass.

(Aside: I also had to write this wrapper to get go-git working on Plan 9:
https://github.com/go-git/go-billy/blob/v5.0.0/osfs/os_plan9.go#L27
but I notice few issues with that one.)

Fixes #86

### DIFF
--- a/convert.go
+++ b/convert.go
@@ -156,7 +156,7 @@ func Move(oldPath string, newPath string, out io.Writer) error {
 			// else we found something unexpected, so to be safe just move it
 			log.Warnw("found unexpected file in datastore directory, moving anyways", "file", fn)
 			newPath := filepath.Join(newDS.path, fn)
-			err := os.Rename(oldPath, newPath)
+			err := rename(oldPath, newPath)
 			if err != nil {
 				return err
 			}
@@ -177,7 +177,7 @@ func moveKey(oldDS *Datastore, newDS *Datastore, key datastore.Key) error {
 	if err != nil && !os.IsExist(err) {
 		return err
 	}
-	err = os.Rename(oldPath, newPath)
+	err = rename(oldPath, newPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


willscott:
I didn't look particularly carefully at this one yet, but the logic seems fine.
I am wondering if maybe this is impetus for getting tests to run in CI for this repo, so that we don't accidentally break plan9.

I'm not fully sure if any of the CIs IPFS has been using have support for plan9 runners. that's probably an assumption to check first before seeing if that's worth doing.

fhs:
I don't think any of the CIs IPFS uses supports Plan 9. I would be happy with just checking if cross compiling to `GOOS=plan9` works. We can do that in the main go-ipfs repo once few more fixes (coming up) lands.

willscott:
to confirm though, cross compiling `GOOS=plan9` would not catch this issue, which would be caught by the test suite, but not by compilation alone.

fhs:
> to confirm though, cross compiling `GOOS=plan9` would not catch this issue, which would be caught by the test suite, but not by compilation alone.

Correct. I don't see any easy way of running the tests. The only public Plan 9 CI I know of is the one from [sr.ht](https://sourcehut.org/blog/2020-05-11-sourcehut-plus-plan-9/) but I never got it to work correctly. The Go project has a custom solution: https://farmer.golang.org/builders

To clarify, I originally found out about the `os.Rename` issue when `ipfs init` was failing. When things are as broken as it is now on Plan 9, we'll find the issues without running the tests. That is of course not a good reason to not set up CI if we can.

willscott:
agreed - i also don't see a great way to automate keeping this from breaking in the future. Your plan seems reasonable for now.

PR #98, opened 2021-11-10
feat: add contexts on datastore methods

guseggert:


convert.go
```diff
@@ -59,7 +60,7 @@ func Move(oldPath string, newPath string, out io.Writer) error {
 	}
 	newDS.deactivate()
```
petar, line 62: Move() should probably take a context eventually. Fine for now.

petar (approved):

convert.go
guseggert: yeah I didn't bother since it's only used by the CLI

PR #100, opened 2021-12-03
fix: add context to DiskUsage()

guseggert:


flatfs.go
```diff
@@ -97,6 +97,7 @@ func combineAccuracy(a, b initAccuracy) initAccuracy {
 }
 
 var _ datastore.Datastore = (*Datastore)(nil)
+var _ datastore.PersistentDatastore = (*Datastore)(nil)
```
aschmahmann, line 100: We need Batching here too, right? Do we also need any type checks on the batch type?

aschmahmann (approved):

PR #105, opened 2022-04-27
Comments, documentations and questions

MichaelMure:
While I was trying to understand that code, I added some comment and documentation to make that process easier for the next person.

I also found a possible bug (https://github.com/ipfs/go-ds-flatfs/commit/26abf67dbdb624fa313b7fcd7d26a17ee3c18b89). It looks like a mistake as `fs` here can't be nil.

Additionally, I'm confused about certain things or noticed some possible improvement (https://github.com/ipfs/go-ds-flatfs/commit/846fdf9a6fa8caf7784649a458d6f8d72cb87c20). I'm using that PR as a way to discuss those points, this last commit can be removed after that.

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -48,7 +48,7 @@ var (
 	// calculating the DiskUsage upon a start when no
 	// DiskUsageFile is present.
 	// If this period did not suffice to read the size of the datastore,
-	// the remaining sizes will be stimated.
+	// the remaining sizes will be estimated.
 	DiskUsageCalcTimeout = 5 * time.Minute
 	// RetryDelay is a timeout for a backoff on retrying operations
 	// that fail due to transient errors like too many file descriptors open.
@@ -118,7 +118,7 @@ func init() {
 // write operations to the same key. See the explanation in
 // Put().
 type Datastore struct {
-	// atmoic operations should always be used with diskUsage.
+	// atomic operations should always be used with diskUsage.
 	// Must be first in struct to ensure correct alignment
 	// (see https://golang.org/pkg/sync/atomic/#pkg-note-BUG)
 	diskUsage int64
@@ -129,7 +129,7 @@ type Datastore struct {
 	shardStr string
 	getDir   ShardFunc
 
-	// sychronize all writes and directory changes for added safety
+	// synchronize all writes and directory changes for added safety
 	sync bool
 
 	// these values should only be used during internalization or
@@ -167,6 +167,8 @@ type op struct {
 	v    []byte        // value
 }
 
+// opMap is a synchronisation structure where a single op can be stored
+// for each key.
 type opMap struct {
 	ops sync.Map
 }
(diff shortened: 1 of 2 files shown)


readme.go
```diff
@@ -1,5 +1,8 @@
 package flatfs
 
+// TODO: now that datastore don't store CIDs but multihashes instead, this is really
```
MichaelMure, line 3: Found out later that it's addressed in https://github.com/ipfs/go-ds-flatfs/issues/103

flatfs.go
```diff
@@ -606,6 +624,12 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
 		if _, err := tmp.Write(value); err != nil {
 			return err
 		}
+
```
MichaelMure, line 627: partially addressed in https://github.com/ipfs/go-ds-flatfs/issues/36

flatfs.go
```diff
@@ -179,7 +181,14 @@ type opResult struct {
...
 
-// Returns nil if there's nothing to do.
+// Begins starts the processing of an op:
+// - if no other op for the same key exist, register it and return immediately
+// - if another op exist for the same key, wait until it's done:
+// 		- if that previous op succeeded, consider that ours shouldn't execute and return nil
+// 		- if that previous op failed, start ours
+//
+// TODO: honestly those rules are weird to me and might lead to data loss? Why a later concurrent op
+//  should not execute?
```
aschmahmann, line 191: I suspect this has to do with flatfs not really being used as a generic datastore but as a backing for a blockstore. This means that (key, value) pairs are immutable so what's your other operation going to do other than a) write the data that's already present b) delete the data you just wrote which is likely due to an application layer bug.

I suspect this isn't the only place where we have these assumptions another is related to the domain of valid key names.

MichaelMure: Something non-obvious when reading from this PR: this synch/dedup structure is shared by all write functions of the datastore, so a `Delete` could be cancelled by a recent `Put` for example. That would be a very legal scenario, even if attached to a blockstore. My point is that this code brings a lot of complexity for a rare case (dedup writes on the same key) for little benefit.

flatfs.go
```diff
@@ -606,6 +624,12 @@ func (fs *Datastore) putMany(data map[datastore.Key][]byte) error {
 		if _, err := tmp.Write(value); err != nil {
 			return err
 		}
+
+		// TODO: I don't understand why the temp files are not closed immediately
+		//  here. After this point we either rename or delete them so why wait?
+		//  It induces the complexity of the closer to avoid having too much open
+		//  files.
+		// TODO: concurrent writes?
```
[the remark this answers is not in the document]
MichaelMure: > Closing the tmp files will result in the data being synced to disk

Ha right, that's a very good point. So that way it's up to the kernel to stream that data on disk however makes the more sense.

MichaelMure:
I think the questions have been discussed enough, we can remove the last commit and merge the rest.

PR #106, opened 2022-05-09
query: also teard down on ctx done

MichaelMure:
I'm not entirely sure this should be supported but at the moment you could expect to terminate a query:
- call `Close()` on the  `Result`
- cancel the context

The second one though doesn't work as cancelling the context doesn't kill the query. This PR address that problem.

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -766,7 +766,7 @@ func (fs *Datastore) Query(ctx context.Context, q query.Query) (query.Results, e
 	// to `Close`.
 	b := query.NewResultBuilder(q)
 	b.Process.Go(func(p goprocess.Process) {
-		err := fs.walkTopLevel(fs.path, b)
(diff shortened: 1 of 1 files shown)


guseggert:
Have you gone through where this is (potentially) used to make sure that returning `context.Canceled` or `context.DeadlineExceeded` are handled gracefully/correctly upstream? Would it make sense to just swallow ctx.Err() and return nothing? (which is what happens when results.Close() is called)

MichaelMure:
My thinking is that calling `results.Close()` is a voluntary action from the caller side.

On the other hand, killing a query with the context is typically done higher in the call stack, with a timeout or linked to a request lifecycle. There, the caller side might not know and an error would help distinguish a success or failure termination.

guseggert:
I did some spot checking of how these bubble up into `go-ipfs` and it looks okay to me. Thanks!

guseggert (approved):

PR #108, opened 2022-05-11
test: faster TestNoCluster by batching the 3200 Puts

MichaelMure:
In normal condition, this makes no difference as the complete test is pretty fast (~200ms), but
as soon as sync is enabled, this commit cut down the duration from 14s to 7s, on a fast SSD.
When running on a slower file system, this make a significant difference.


MichaelMure:
I did the same thing with `TestQueryLeak` for the same reason.

PR #109, opened 2022-08-02
fix: actually use the size hint in util_windows.go

Jorropo:
This code was broken, it computed the size hint to do nothing with it.

Found with this profile:
Capture d’écran du 2022-08-02 15-40-33

This profile clearly show a pattern of `append`'s growth.
With this new code the only way to `append` is if the size hint undershoot (which would require more data to be written between the `Size` and `Read` calls). Might be worth a `panic("race")` or returning an error if that happen ?

### DIFF
--- a/util_windows.go
+++ b/util_windows.go
@@ -12,7 +12,7 @@ package flatfs
 
 import (
 	"bytes"
-	"io"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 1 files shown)


Jorropo:
Found with @Arlodotexe

util_windows.go
```diff
@@ -93,10 +93,30 @@ func readFileOnce(filename string) ([]byte, error) {
...
+				break
+			}
+			return nil, err
+		}
+
+		// Size hint undershoot, let's grow the buffer
+		if len(remaining) == 0 {
+			dataSize := len(data)
+			data = append(data, 0) // use append's growth pattern
+			remaining = data[dataSize:]
```
Stebalien, line 117: I'd just re-slice each time (f.Read(`data[total:])`) to simplify, rather than keeping two buffers.

util_windows.go
```diff
@@ -93,10 +93,30 @@ func readFileOnce(filename string) ([]byte, error) {
...
+			if err == io.EOF {
+				break
+			}
+			return nil, err
+		}
+
+		// Size hint undershoot, let's grow the buffer
+		if len(remaining) == 0 {
+			dataSize := len(data)
+			data = append(data, 0) // use append's growth pattern
```
Stebalien, line 116: This will read one byte at a time unless you update the capacity.

Anyways, I'd just explicitly expand by 2x and make it obvious.

util_windows.go
```diff
@@ -93,10 +93,30 @@ func readFileOnce(filename string) ([]byte, error) {
...
 		// it just right.
-		if size := fi.Size() + bytes.MinRead; size > n {
-			n = size
+		n += fi.Size()
+	}
+
+	data := make([]byte, n)
+	remaining := data
+	var total int64
+	for {
```
Stebalien, line 102: You can `io.Copy` into a `bytes.Buffer` instead. That'll handle growing and everything else. `bytes.Buffer` implements `ReadFrom` with code that's pretty much identical to the code you have here: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/bytes/buffer.go;l=198-215

Jorropo: Yeah mb, I've miss red the docs, I've seen `unred` and thought this returned the extra capacity, but it's infact red from the buffer's point of view, not the file's one.

Stebalien (approved):

Jorropo:
I overlooked it didn't built ... :'(
I'll release after I get an `O_TMPFILE` patch ready for linux.

PR #118, opened 2024-01-25
feat: renterd backed flatfs

LexLuthr:
This feature connects the flatfs with a renterd backend. All the writes are copied over to Sia network in real time. Any failed operation results on failure for IPFS as well. The DELETE operations can be copied over if specified by the user.


LexLuthr:
Incorrect Repo. Closing.

PR #136, opened 2025-08-27
Rewrite batch mode to use temp directory

requilence:
### Problem

 When uploading large files through UnixFS or other IPLD structures, the current batch implementation keeps everything in memory and writes blocks individually. If the application crashes mid-process, partial blocks remain in FlatFS, making garbage collection extremely difficult since there's no way to identify which blocks belong to incomplete operations.

### Solution

  This PR rewrites the batch mode to use a temporary directory approach:

  1. Atomic batch operations: All blocks in a batch are written to a temporary directory first, then renamed to their final locations only on commit. If the process crashes, no partial data remains in the main datastore - the temp directory is cleaned on restart.
  2. Memory efficiency: Instead of keeping all data in memory, blocks are written directly to temp files, enabling large file uploads without high RAM usage.
  3. Read operations in batch: Added `BatchReader` interface implementing Get/Has/GetSize methods. This is essential for IPLD structures which need to verify block links during construction. Without batch reads, you cannot build complex IPLD structures that reference blocks added earlier in the same batch. This follows standard database transaction semantics where reads within a transaction see uncommitted writes.

### Implementation Details

  - Each batch creates a unique temp directory under .temp/
  - Files are written without sharding in temp
  - On commit, files are atomically renamed to their sharded destinations
  - On discard or crash, temp files are cleaned up (on the next startup). This code was already existing
  - Thread-safe with mutex protection
  - Batch can be reused after commit/discard (because this behavior of Batch was before my changes)

### DIFF
--- a/flatfs.go
+++ b/flatfs.go
@@ -1124,7 +1124,8 @@ const maxConcurrentPuts = 16
 // IMPORTANT: Batch instances should not be reused after Commit or Discard.
 type flatfsBatch struct {
 	mu      sync.Mutex
-	puts    []datastore.Key
+	puts    []datastore.Key            // ordered list for iteration (Commit, Query)
+	putsSet map[datastore.Key]struct{} // O(1) lookup for Get/Has/GetSize
 	deletes map[datastore.Key]struct{}
 
 	ds      *Datastore


gammazero:
Need to review. Check that there is still ability to do in-memory batching for smaller batches.

flatfs.go
```diff
@@ -1210,46 +1083,400 @@ func (fs *Datastore) Close() error {
...
+		if _, err := bt.ds.makeDirNoSync(dir); err != nil {
+			return fmt.Errorf("failed to create directory: %w", err)
+		}
+		dirsToSync[dir] = struct{}{}
+	}
+
+	// Move all temp files to their final destinations
+	for i, key := range bt.puts {
+		fileName := bt.tempFiles[i]
+		tempFile := filepath.Join(bt.tempDir, fileName)
```
gammazero, line 1432:
```suggestion
		tempFile := filepath.Join(bt.tempDir, bt.tempFiles[i])
```

batch_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+func collectQueryResults(t *testing.T, results query.Results) []query.Entry {
+	var entries []query.Entry
+	for result := range results.Next() {
+		if result.Error != nil {
+			t.Fatalf("query result error: %v", result.Error)
+		}
+		entries = append(entries, result.Entry)
+	}
+	return entries
```
gammazero, line 533:
```suggestion
	t.Helper()
	entries, err := results.Rest()
	if err != nil {
		t.Fatalf("query result error: %v", result.Error)
	}
	return entries
```

flatfs.go
```diff
@@ -1210,46 +1083,400 @@ func (fs *Datastore) Close() error {
...
 	if !keyIsValid(key) {
 		return fmt.Errorf("when putting '%q': %v", key, ErrInvalidKey)
 	}
-	bt.puts[key] = val
+
+	bt.mu.Lock()
+	defer bt.mu.Unlock()
+
+	// Check if batch temp directory exists, recreate if needed (for reuse after commit/discard)
+	if _, err := os.Stat(bt.tempDir); os.IsNotExist(err) {
```
gammazero, line 1133: New code should use `errors.Is(err, fs.ErrNotExist)` instead of `os.IsNotExist(err)`. See: https://pkg.go.dev/os@go1.25.1#IsNotExist
```suggestion
	_, err := os.Stat(bt.tempDir)
	if errors.Is(err, fs.ErrNotExist) {
```

flatfs.go
```diff
@@ -1210,46 +1083,400 @@ func (fs *Datastore) Close() error {
...
+		if err := os.Mkdir(bt.tempDir, 0755); err != nil {
+			return fmt.Errorf("failed to recreate batch temp directory: %w", err)
+		}
+	}
```
gammazero, line 1137: Should probably handle non-nil error if directory exists but `stat` fails (permissions problem), instead of doing more work only to have it fail when trying to write to it.
```suggestion
	} else if err != nil {
		return fmt.Errorf("cannot stat batch temp directory: %w", err)
	}
```

gammazero (changes requested):
My concern is that this will negatively impact the performance of `Put. This will make `Commit` faster because the temp files are already written, but existing code my expect a `Put` that is not slowed down by writing to the filesystem.

It may be worth considering having `Put` write the temp files asynchronously in a separate goroutine. Then `Commit` will wait for any outstanding `Put` to finish, handle any errors from async `Put`s , and then move all the temp files to their final destination.

hsanjuan:
Triage questions that come up:

* Can you describe the scenario in which memory usage of batches becomes a problem? Is this while using Kubo or are you using flatfs and interacting with it in a different application?

requilence:
> Triage questions that come up:
> 
> * Can you describe the scenario in which memory usage of batches becomes a problem? Is this while using Kubo, or are you using flatfs and interacting with it in a different application?

We are using flatfs and the IPLD structures in Anytype (http://github.com/anyproto/anytype-heart) to store files on users' devices. If we are not using a batch operation, we can find ourselves in a situation where the user force-closes the app in the middle of a large file (e.g., 4 GB video) addin, which will leave a lot of garbage blocks in flatfs. Therefore, we are using atomic batch operations. At the same time, we are trying to avoid heap allocations and reuse memory, as it is especially critical on mobile devices.

@lidel, I'm not sure async Put is really a good way for us. We would rather do it synchronously, otherwise it could lead to heap spikes.

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
 	deletes map[datastore.Key]struct{}
 
-	ds *Datastore
+	ds      *Datastore
+	tempDir string
+
+	// Async write tracking
+	asyncWrites     sync.WaitGroup
+	asyncMu         sync.Mutex
+	asyncFirstError error
```
gammazero:
```suggestion
	asyncFirstError error
	asyncPutGate    chan struct{}
```

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
+	datastore.Batch
+	Discard(ctx context.Context) error
+}
+
+// BatchReader is an optional interface for batches that support read operations
+type BatchReader interface {
+	datastore.Batch
+	datastore.Read
+}
+
```
gammazero:
```suggestion

// TODO: move this to be with other consts above
const maxConcurrentPuts = 16

```

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
+	tempDir := filepath.Join(fs.tempPath, fmt.Sprintf("batch-%d-%d", time.Now().UnixNano(), rand.Int63()))
+	if err := os.Mkdir(tempDir, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create batch temp directory: %w", err)
+	}
+
 	return &flatfsBatch{
-		puts:    make(map[datastore.Key][]byte),
 		deletes: make(map[datastore.Key]struct{}),
 		ds:      fs,
+		tempDir: tempDir,
```
gammazero:
```suggestion
		deletes:      make(map[datastore.Key]struct{}),
		ds:           fs,
		tempDir:      tempDir,
		asyncPutGate: make(chan struct{}, maxConcurrentPuts),
```

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
 		return fmt.Errorf("when putting '%q': %v", key, ErrInvalidKey)
 	}
-	bt.puts[key] = val
+
+	if err := bt.getAsyncError(); err != nil {
+		// If there was an error from a previous async write, return it fast
+		// This may be useful if, for example, we are out of disk space.
+		return err
+	}
+
```
gammazero:
```suggestion

	// Acquire semaphore slot before starting another async put.
	select {
	case bt.asyncPutGate <- struct{}{}:
	case <-ctx.Done():
		return ctx.Err()
	}

```

gammazero (changes requested):
See and commit suggestions: Added a semaphore to limit concurrent puts

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
 	if keyIsValid(key) {
 		bt.deletes[key] = struct{}{}
 	} // otherwise, delete is a no-op anyways.
 	return nil
 }
 
+func (bt *flatfsBatch) Get(ctx context.Context, key datastore.Key) ([]byte, error) {
```
lidel: Is this accurate?
```suggestion
// Get retrieves a value from the batch or underlying datastore.
//
// Transaction semantics: Returns uncommitted data written to this batch via Put,
// even before Commit. This allows building IPLD structures that reference blocks
// added earlier in the same batch.
//
// Performance: O(n) where n is the number of Put operations, as it must scan
// the puts slice to check if the key exists in the batch.
func (bt *flatfsBatch) Get(ctx context.Context, key datastore.Key) ([]byte, error) {
```

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
+			return -1, err
+		}
+		return int(stat.Size()), nil
+	}
+
+	// If not in batch, check main datastore
+	return bt.ds.GetSize(ctx, key)
+}
+
+func (bt *flatfsBatch) Query(ctx context.Context, q query.Query) (query.Results, error) {
```
lidel: Is this accurate description?
```suggestion
// Query returns all entries from both the batch and underlying datastore,
// properly merging results to reflect the batch's uncommitted state.
//
// Merge logic:
// - Keys written via Put appear in results (even if not committed)
// - Keys marked for Delete do not appear in results
// - Keys Put multiple times appear only once (last write wins)
// - Main datastore results are excluded if overwritten or deleted in batch
//
// The implementation waits for all async writes to complete before querying.
func (bt *flatfsBatch) Query(ctx context.Context, q query.Query) (query.Results, error) {
```

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
+	// Remove the batch temp directory and all subdirectories
+	_ = os.RemoveAll(bt.tempDir)
+	bt.puts = nil
+	bt.asyncFirstError = nil
+	bt.deletes = make(map[datastore.Key]struct{})
+
+	return nil
+}
+
 func (bt *flatfsBatch) Commit(ctx context.Context) error {
```
lidel: Same here
```suggestion
// Commit atomically applies all batch operations to the datastore.
//
// Atomicity guarantee: All Put operations are moved to their final destinations
// only after being written to temp files. If the process crashes before Commit
// completes, the temp directory is cleaned on restart and no partial data remains.
//
// Order of operations:
// 1. Wait for all async Put goroutines to complete
// 2. Check for any async write errors (fail-fast)
// 3. Create all destination shard directories
// 4. Atomically rename temp files to final sharded paths
// 5. Apply all Delete operations
// 6. Sync directories (if sync is enabled)
//
// Concurrency: Uses doWriteOp to handle concurrent commits gracefully.
// If another goroutine commits the same key, the operation succeeds.
func (bt *flatfsBatch) Commit(ctx context.Context) error {
```

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
+			os.Remove(tempFile)
+			bt.setAsyncError(fmt.Errorf("failed to close temp file: %w", err))
+			return
+		}
+	}(val)
+
 	return nil
 }
 
+// setAsyncError save the first error from an async write operation if it is not already set.
```
lidel:
```suggestion
// setAsyncError saves the first error from an async write operation.
// Only the first error is captured; subsequent errors are ignored.
// This provides fail-fast behavior: once any write fails, subsequent
// operations return that error immediately.
```

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
+	tempDir string
+
+	// Async write tracking
+	asyncWrites     sync.WaitGroup
+	asyncMu         sync.Mutex
+	asyncFirstError error
 }
 
 func (fs *Datastore) Batch(_ context.Context) (datastore.Batch, error) {
+	// Create a unique temp directory for this batch
```
lidel:
```suggestion
	// Create a unique temp directory for this batch
  // Note: Temp files are not sharded (flat structure) for simplicity and speed.
  // Files are only sharded when renamed to their final destination on Commit.
```

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
+	Discard(ctx context.Context) error
+}
+
+// BatchReader is an optional interface for batches that support read operations
+type BatchReader interface {
+	datastore.Batch
+	datastore.Read
+}
+
 type flatfsBatch struct {
```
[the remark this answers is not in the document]
gammazero: Yes, this is correct AFAIK.

flatfs.go
```diff
@@ -1210,46 +1086,452 @@ func (fs *Datastore) Close() error {
...
 	return &flatfsBatch{
-		puts:    make(map[datastore.Key][]byte),
 		deletes: make(map[datastore.Key]struct{}),
 		ds:      fs,
+		tempDir: tempDir,
 	}, nil
 }
 
+// Put writes val for key to a temporary file asynchronously and returns immediately.
+// The caller must not modify or reuse val
```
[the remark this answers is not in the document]
gammazero: Note: This requirement also existed in the previous `flatfsBatch.Put` code where the value being written was stored in memory and reuse of the `val` would corrupt data.

PR #150, opened 2026-06-24
fix: make last batch operation take effect

### LINKED ISSUE #49: Handle puts *and* deletes in Batch
See: https://github.com/ipfs/go-datastore/pull/103. Either that or we need to document the semantics of batch (and say that it first puts, then deletes).

gammazero:
When adding Put and Delete operations to a batch, for the same key, make the last operation be the one to take effect.

Put and Delete for same key:
- Delete after Put: Commit removes Put and deletes item from datastore.
- Put after Delete: Put removes Delete from batch and adds Put.

Closes #49

### DIFF
--- a/batch_test.go
+++ b/batch_test.go
@@ -344,6 +344,135 @@ func testBatchDiscard(dirFunc mkShardFunc, t *testing.T) {
 	}
 }
 
+func TestBatchPutDeleteSameKey(t *testing.T) {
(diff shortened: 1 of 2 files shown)


lidel (approved):
LGTM, feels correct, thanks for locking this down with docs and test (pushed small typo fix).

